PR #1, opened 2018-06-14
tw/develop

twittner:


src/frame/codec.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+impl Encoder for FrameCodec {
+    type Item = RawFrame;
+    type Error = io::Error;
+
+    fn encode(&mut self, frame: Self::Item, bytes: &mut BytesMut) -> Result<(), Self::Error> {
+        self.header_codec.encode(frame.header, bytes)?;
+        self.body_codec.encode(frame.body.0, bytes)
```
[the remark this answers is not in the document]
twittner: This should not be problematic. `FramedWrite`, which uses `Encoder`, assumes that when an error is returned by `encode` that `BytesMut` is  modified in arbitrary ways and not safe to use. Cf. https://docs.rs/tokio-io/0.1.7/src/tokio_io/_tokio_codec/framed_write.rs.html#178

twittner:
AFAIK this would actually be MIT. CC @tomaka

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+impl io::Read for Stream {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if self.state == State::Closed || self.state == State::RecvClosed {
+            return Ok(0)
+        }
+        if self.poll_receiver().is_ready() {
+            return Ok(0)
+        }
+        if self.buffer.is_empty() {
+            return Err(io::Error::new(io::ErrorKind::WouldBlock, "not ready"))
```
tomaka: I'd prefer a `match` block, so that we clearly associate `Async::NotReady` to `WouldBlock`.
It is a logic error to return `WouldBlock` if a `NotReady` has not generated, so I think it's a good idea to clearly show the association in the source code.

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+                }
+            }
+        }
+    }
+}
+
+impl io::Read for Stream {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        if self.state == State::Closed || self.state == State::RecvClosed {
+            return Ok(0)
```
tomaka: That means that if a node sends some data then closes the stream immediately after, there's a change that we won't receive the data?

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+pub struct Stream {
+    id: Id,
+    state: State,
+    config: Arc<Config>,
+    recv_window: Arc<Window>,
+    send_window: u32,
+    buffer: BytesMut,
+    sender: Sender,
+    receiver: Receiver,
+    writer_task: Option<Task>
```
tomaka: Shouldn't that be a `Vec` (or something similar) in case multiple different tasks try to write "simultaneously"?

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+                                error!("[{}] failed to send window update: {}", self.id, e);
+                                self.state = State::Closed;
+                            }
+                            self.recv_window.set(self.config.receive_window as usize);
+                        }
+                        continue
+                    }
+                    Some(Item::WindowUpdate(n)) => {
+                        trace!("[{}] received window update: {}", self.id, n);
+                        self.send_window = self.send_window.checked_add(n).unwrap_or(u32::MAX);
```
tomaka: `saturating_add`

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        if self.send_window == 0 {
+            trace!("[{}] write: send window exhausted", self.id);
+            self.writer_task = Some(task::current());
+            return Err(io::Error::new(io::ErrorKind::WouldBlock, "window empty"))
+        }
+
+        let len = min(buf.len(), self.send_window as usize);
+        self.send_window -= len as u32;
```
tomaka: `Body::from_bytes` checks that `len <= u32::max()`, but that happens *after* this line where you do `len as u32`.
There's no chance that we try to send 4 GB of data I think, but I think the code is a bit absurd as it is.
Maybe do `Body::from_bytes(...).expect(...)` instead?

src/connection.rs
```diff
@@ -0,0 +1,462 @@
...
+
+impl<T> Connection<T>
+where
+    T: AsyncRead + AsyncWrite
+{
+    /// Create a new connection either in client or server mode.
+    pub fn new(resource: T, config: Arc<Config>, mode: Mode) -> Self {
+        debug!("new connection");
+        let controller = {
+            let (tx, rx) = mpsc::channel(1024);
```
tomaka: Isn't `unbounded` better?

src/connection.rs
```diff
@@ -0,0 +1,462 @@
...
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if self.is_dead {
+            return Ok(Async::Ready(None))
+        }
+
+        // First, check for pending frames we need to send.
+        if let Some(frame) = self.pending.take() {
+            trace!("send pending: {:?}", frame);
+            try_ready!(self.send(frame))
```
tomaka: If this returns `NotReady`, `self.pending` will be empty at the next iteration.

src/connection.rs
```diff
@@ -0,0 +1,462 @@
...
+        self.is_dead = true;
+        self.streams.clear()
+    }
+
+    fn send(&mut self, frame: RawFrame) -> Poll<(), ConnectionError> {
+        trace!("send: {:?}", frame);
+        match self.resource.start_send(frame) {
+            Ok(AsyncSink::Ready) => Ok(Async::Ready(())),
+            Ok(AsyncSink::NotReady(frame)) => {
+                self.pending = Some(frame);
```
tomaka: I think this should have a `debug_assert` that `self.pending` is not `None`, so that we can possible bugs.

src/stream.rs
```diff
@@ -0,0 +1,301 @@
...
+                    return Async::Ready(())
+                }
+                Ok(Async::NotReady) => {
+                    return Async::NotReady
+                }
+                Ok(Async::Ready(item)) => match item {
+                    Some(Item::Data(body)) => {
+                        trace!("[{}] received data: {:?}", self.id, body);
+                        let body_len = body.bytes().len();
+                        self.buffer.extend(body.into_bytes());
```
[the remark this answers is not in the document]
twittner: Yes. The idea is that the connection checks the stream's receive window before delivering a data item to the stream. The window is accessible from the connection (cf. https://github.com/paritytech/yamux/blob/tw/develop/src/connection.rs#L79), but I did not yet get around to implement the check.

src/connection.rs
tomaka: Oh nevermind, it will set `self.pending` back to the same frame.

src/stream.rs
twittner: `self.send_window` is a `u32` so `len` can not exceed `u32::MAX`.

src/connection.rs
twittner: No, because `Connection::send` puts its argument back into `Connection::pending` before returning `NotReady`.

twittner: Sorry for the comment. GitHub preferred to not show me your second comment.

src/connection.rs
twittner: Why do you think so? I thought it would maybe be a good idea to exert some back-pressure if we are late opening a certain number of streams.

PR #3, opened 2018-07-05
 Refactor to not use mpsc channels.

twittner:


src/lib.rs
```diff
@@ -30,27 +31,51 @@ extern crate tokio_io;
...
+            receive_window: DEFAULT_CREDIT,
+            max_buffer_size: 1024 * 1024,
+            max_num_streams: 8192
         }
     }
 }
 
+impl Config {
+    #[must_use]
+    pub fn set_receive_window(&mut self, n: u32) -> bool {
```
tomaka: Should be a `Result`.

src/connection.rs
```diff
@@ -17,484 +17,505 @@
...
 
-    /// Get a control handle which allows to open new streams.
-    pub fn control(&self) -> Ctrl {
-        self.controller.sender.clone()
+    fn next_stream_id(&mut self) -> Result<stream::Id, ConnectionError> {
+        if self.next_id > u32::MAX - 2 {
+            return Err(ConnectionError::NoMoreStreamIds)
+        }
+        let proposed = stream::Id::new(self.next_id);
+        self.next_id += 2;
```
tomaka: I'd suggest using `checked_add(2)` here, and return `ConnectionError::NoMoreStreamIds` if it fails, and remove the `if` above.

src/connection.rs
```diff
@@ -17,484 +17,505 @@
...
+            task.notify()
+        }
+    }
+
+    fn flush_pending(&mut self) -> Poll<(), ConnectionError> {
+        while let Some(frame) = self.pending.pop_front() {
+            trace!("{:?}: send: {:?}", self.mode, frame.header);
+            if let AsyncSink::NotReady(frame) = self.resource.start_send(frame)? {
+                self.pending.push_front(frame);
+                try_ready!(self.resource.poll_complete())
```
tomaka: I think this line should be `return Async::NotReady;`.

The documentation of `start_send` mentions that calling `poll_complete()` here is useless:
> The method returns AsyncSink::NotReady if the sink was unable to begin sending, usually due to being full. The sink must have attempted to complete processing any outstanding requests (equivalent to poll_complete) before yielding this result.

Also, I guess there's a possibility that `start_send` returns `NotReady` and `poll_complete()` returns `Ready`, in which case this will loop forever.

src/connection.rs
```diff
@@ -17,484 +17,505 @@
...
+            }
+        }
+        try_ready!(self.resource.poll_complete());
+        self.wake_tasks();
```
tomaka: This seems to be here for no specific reason :sweat_smile:

PR #22, opened 2018-10-29
Prepare release

twittner:
Addresses https://github.com/libp2p/rust-libp2p/issues/593.

cc: @gnunicorn, @folsen


tomaka:
The only question is whether the license change to Apache/MIT corresponds to what we want.

src/connection.rs
```diff
@@ -72,6 +67,11 @@ where
         }
     }
 
+    /// Open a new outbound stream which is multiplexed over the existing connection.
```
dvdplm: Maybe s/Open/Opens/

folsen:
> The only question is whether the license change to Apache/MIT corresponds to what we want.

MIT/Apache seems strictly better than pure MIT in my view. It also aligns better with Substrate which will be Apache-2.

5chdn:
Is there a reason to still maintain MIT?

twittner:
I would argue similarly to this: https://www.rust-lang.org/en-US/faq.html#why-a-dual-mit-asl2-license

PR #24, opened 2018-12-03
Add inherent poll() to Connection

dvdplm:
`Stream::poll()` takes a `&mut self` which is problematic as libp2p's `StreamMuxer`s methods take immutable references. This was previously worked around in libp2p by wrapping `yamux::Connection` in an extra `Mutex`. This PR implements `poll()` as an inherent method that takes `&self` making it possible to call without a mutex from `libp2p`.
The PR retains the `Stream` impl so that code can use its convenient API.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -133,6 +133,29 @@ where
         connection.on_drop(Action::None);
         Ok(result)
     }
+
+    /// Poll connection for incoming data
+    pub fn poll(&self) -> Poll<Option<StreamHandle<T>>, ConnectionError> {
+        let mut connection = Use::with(self.inner.lock(), Action::Destroy);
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -156,11 +150,23 @@ where
...
                     continue
                 };
-                connection.on_drop(Action::None);
-                return Ok(Async::Ready(Some(stream)))
+            connection.on_drop(Action::None);
+            return Ok(Async::Ready(Some(stream)))
         }
         connection.on_drop(Action::None);
-        Ok(Async::NotReady)
+        return Ok(Async::NotReady)
```
twittner:
```suggestion
        Ok(Async::NotReady)
```

PR #25, opened 2018-12-06
Fix problem where a fast client can "drown" a server

dvdplm:
When a `yamux::Connection` is setting up a new incoming connection ([`process_incoming()`](https://github.com/paritytech/yamux/blob/master/src/connection.rs#L146)) it tries to make as much progress as possible. 
If a client connection is fast enough to send all its data and then closes the connection, the `Connection` reads all the data within a single call to `process_incoming()`, including the end of the stream. This means that [it will set `is_dead` to true](https://github.com/paritytech/yamux/blob/master/src/connection.rs#L337) on the connection and clear the stream, which will cause `poll()` to [return `Ready(None)`](https://github.com/paritytech/yamux/blob/master/src/connection.rs#L148) which bubbles up to the muxer where it is interpreted as if the connection was gone. The result is that `muxing::inbound_from_ref_and_wrap()` [returns `None`](https://github.com/libp2p/rust-libp2p/blob/master/core/src/muxing.rs#L176) and the data read is lost.
This PR re-orders things such that `poll()` first checks for incoming connections and then check for a dead connection and clears the stream if needed.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -144,9 +144,6 @@ where
     pub fn poll(&self) -> Poll<Option<StreamHandle<T>>, ConnectionError> {
         let mut connection = Use::with(self.inner.lock(), Action::Destroy);
         connection.process_incoming()?;
-        if connection.is_dead {
-            return Ok(Async::Ready(None))
-        }
 
         while let Some(id) = connection.incoming.pop_front() {
             let stream =
@@ -159,6 +156,12 @@ where
             connection.on_drop(Action::None);
             return Ok(Async::Ready(Some(stream)))
         }
+
+        if connection.is_dead {
+            connection.streams.clear();
+            return Ok(Async::Ready(None))
+        }
+
         connection.on_drop(Action::None);
         Ok(Async::NotReady)
     }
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -497,6 +499,7 @@ where
 }
 
 /// A handle to a multiplexed stream.
+#[derive(Debug)]
```
dvdplm: This has nothing to do with the PR subject. Ok to keep it here anyway?

src/connection.rs
```diff
@@ -159,6 +156,12 @@ where
             connection.on_drop(Action::None);
             return Ok(Async::Ready(Some(stream)))
         }
+
+        if connection.is_dead {
+            connection.streams.clear();
```
dvdplm: Should the `notify_all()` call be here or where it was originally?

twittner:
I agree that short lived connections that come and go within a single `poll` call may not produce any incoming streams. To handle this properly I have made some more changes, which are the subject of PR #26.

src/connection.rs
dvdplm: I guess this is the only thing added by this PR! :D

twittner: Forgot about this one ;-). Have added it to #26.

dvdplm:
Closing in favour of #26

PR #26, opened 2018-12-06
No clear

twittner:
To allow streams to finish reading all of their data as well as to allow short-lived connections which start and finish within a single `poll` call, we should never clear the whole streams map, but rather have streams remove themselves one by one.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -124,7 +124,7 @@ where
         }
         let result = {
             let c = &mut *connection;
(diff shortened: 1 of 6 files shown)


src/connection.rs
```diff
@@ -496,6 +496,7 @@ where
 }
 
 /// A handle to a multiplexed stream.
+#[derive(Debug)]
```
dvdplm: 👍

PR #29, opened 2018-12-19
Make config closer to mplex

dvdplm:
Rernames `max_num_streams` to `max_substreams` ([like mplex](https://github.com/libp2p/rust-libp2p/blob/master/muxers/mplex/src/lib.rs#L44)) and adds `Config::new()` ([mplex has this; uses `Default::default()`](https://github.com/libp2p/rust-libp2p/blob/master/muxers/mplex/src/lib.rs#L56)).

I'm not sure this is a good idea or not. I think it'd be good if different `StreamMuxer` impls try to be as similar as makes sense, to make it easy on third party devs to learn their APIs.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -83,7 +83,7 @@ where
         if connection.is_dead {
             return Ok(None)
         }
-        if connection.streams.len() >= connection.config.max_num_streams {
+        if connection.streams.len() >= connection.config.max_substreams {
             error!("maximum number of streams reached");
             return Err(ConnectionError::TooManyStreams)
         }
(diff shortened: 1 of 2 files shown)


twittner:
I do not see why there should be a need for aligning `StreamMuxer` implementations. The point of that trait is to ensure there exists a common interface. Beyond that there are inevitable differences between muxers.

src/connection.rs
```diff
@@ -83,7 +83,7 @@ where
         if connection.is_dead {
             return Ok(None)
         }
-        if connection.streams.len() >= connection.config.max_num_streams {
+        if connection.streams.len() >= connection.config.max_substreams {
```
twittner: Within this crate the notions used are `Connection` and `Stream`. The sudden use of the term "substream" in configuration is alien to that, so I would prefer to not make this change.

dvdplm:
> I do not see why there should be a need for aligning `StreamMuxer` implementations.

That is a valid point of view. I agree it is not **necessary**, but I do think it's useful if they are.

Going to close this unless @tomaka has a strong opinion.

tomaka:
My opinion is that it's better if the `yamux` crate exposes yamux-specific language and concepts, and is not designed specifically for libp2p.

PR #31, opened 2018-12-20
Always attempt to flush data in `process_incoming`.

twittner:
Currently, if `poll_flush_notify` returns `NotReady` in `flush_pending` after we managed to start sending our pending data we will not attempt to resume flushing until more pending data has been accumulated which does not necessarily happen.

Instead, we must attempt to flush unconditionally in `process_incoming` to ensure any pending flush operation completes.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -294,7 +294,6 @@ where
         if self.is_dead {
             return Ok(Async::Ready(()))
         }
-        try_ready!(self.resource.poll_flush_notify(&self.tasks, 0));
         while let Some(frame) = self.pending.pop_front() {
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -311,7 +310,8 @@ where
             return Ok(Async::Ready(()))
         }
         loop {
-            if !self.pending.is_empty() && self.flush_pending()?.is_not_ready() {
+            try_ready!(self.resource.poll_flush_notify(&self.tasks, 0));
+            if !self.pending.is_empty() & self.flush_pending()?.is_not_ready() {
```
dvdplm: Did you mean `&&` here?

PR #36, opened 2019-02-11
Check stream status again.

twittner:
`process_incoming` might have received a RESET frame for the reading stream. If this is the only stream and we do not check the status we are waiting for the task to be woken up again, however no more data may be sent over the connection.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -614,7 +614,13 @@ where
                         continue
                     }
                     inner.on_drop(Action::None);
-                    return Err(io::ErrorKind::WouldBlock.into())
+                    let can_read = inner.streams.get(&self.id).map(|s| s.state().can_read());
+                    if can_read.unwrap_or(false) {
+                        return Err(io::ErrorKind::WouldBlock.into())
+                    } else {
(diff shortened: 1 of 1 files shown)


src/connection.rs
```diff
@@ -614,7 +614,13 @@ where
                         continue
                     }
                     inner.on_drop(Action::None);
-                    return Err(io::ErrorKind::WouldBlock.into())
+                    let can_read = inner.streams.get(&self.id).map(|s| s.state().can_read());
+                    if can_read.unwrap_or(false) {
+                        return Err(io::ErrorKind::WouldBlock.into())
+                    } else {
+                        debug!("{}: {}: can no longer read", inner.id, self.id);
+                        return Ok(0) // stream has been reset
```
tomaka: I would say that we should return an error if the stream has been reset, and EOF if it has been closed.

twittner: Maybe this can be postponed? Within this PR we can not distinguish between close and reset and the check should only fix the problem described in the commit message. Once this and #37 are in we can focus on shutdown and FIN and then change this maybe.

PR #38, opened 2019-02-13
Send FIN frame on shutdown.

twittner:
When calling `AsyncWrite::shutdown` on a stream, do not send a RESET frame but a FIN frame. This closes the write end on the sender side but allows for reading data from remote still. The remote will half-close its receiving side of the stream but can still write data. If the remote also shuts down the stream it will be fully closed and no RESET frame will be sent. However half-closed streams still send a RESET frame when dropped.


tomaka:
Some testing would be nice as well eventually.

PR #40, opened 2019-02-14
Randomise and expand integration tests.

romanb:
Includes:

  * Using OS-assigned, free port numbers.
  * Randomising test inputs.
  * Extracting reusable functionality.
  * Begin extending integration test coverage, e.g. for  https://github.com/paritytech/yamux/pull/36. More tests may follow.

Thereby upgrade the `quickcheck` dependency.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ mod stream;
 
 pub use crate::connection::{Connection, Mode, StreamHandle};
 pub use crate::error::{DecodeError, ConnectionError};
+pub use crate::stream::State;
 
 pub(crate) const DEFAULT_CREDIT: u32 = 256 * 1024; // as per yamux specification
 
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -18,7 +18,7 @@ use std::io;
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
 use tokio::{net::{TcpListener, TcpStream}, runtime::Runtime};
 use tokio_codec::{BytesCodec, Framed, Encoder, Decoder};
-use yamux::{ConnectionError, Config, Connection, Mode};
+use yamux::{Config, Connection, Mode};
 
 #[test]
 fn prop_send_receive() {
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -27,6 +27,7 @@ mod stream;
 
 pub use crate::connection::{Connection, Mode, StreamHandle};
 pub use crate::error::{DecodeError, ConnectionError};
+pub use crate::stream::State;
```
romanb: This is simply because I discovered in the API docs that `state(&self) -> Option<State>` is a publicly exported function of `StreamHandle`, but `State` was not visible.

tests/smoke.rs
```diff
@@ -8,97 +8,191 @@
...
-                    error!("C: connection error: {}", e);
-                    Either::B(future::ok(Loop::Break(())))
-                }
+/// Sequentially send a sequence of messages on a connection, with a new
+/// stream for each message, collecting the responses.
+fn loop_send_recv<D,I>(c: Connection<TcpStream>, d: D, i: I)
+    -> impl Future<Item = Vec<<D as Decoder>::Item>, Error = ()>
+where
+    I: Iterator<Item = <D as Encoder>::Item>,
+    D: Encoder + Decoder + Copy,
```
twittner: Maybe `Clone` would be more appropriate than `Copy` for a codec?

tests/smoke.rs
```diff
@@ -8,97 +8,191 @@
...
+    D: Encoder<Error = io::Error> + Decoder<Error = io::Error> + Copy,
+    <D as Encoder>::Item: From<<D as Decoder>::Item>
+{
+    c.for_each(move |stream| {
+        let (stream_out, stream_in) = Framed::new(stream, d).split();
+        stream_in
+            .take(n)
+            .map(|frame_in| frame_in.into())
+            .forward(stream_out)
+            .map_err(|e| ConnectionError::Io(e))
```
twittner:
```suggestion
            .from_err()
```

tests/smoke.rs
```diff
@@ -8,97 +8,191 @@
...
+        let max_streams = n % 100;
+        cfg.set_max_num_streams(max_streams);
+        let codec = BytesCodec::new();
+        let server = server(cfg.clone(), l).and_then(move |c| repeat_echo(c, codec, 1));
+        let client = client(cfg, a).and_then(move |conn| {
+            Ok((0 .. max_streams + 1).fold(vec![], |mut v, _|
+                match conn.open_stream() {
+                    Ok(Some(stream)) => { v.push(stream); v },
+                    _ => v
+                }))
```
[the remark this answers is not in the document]
romanb: I went with a slight modification, as checking the vector length seems redundant with this variant, i.e.:

```rust
let mut v = Vec::new();
for _ in 0 .. max_streams {
    match conn.open_stream() {
        Ok(Some(s)) => v.push(s),
        Ok(None) => panic!("unexpected EOF when opening stream"),
        Err(e) => panic!("unexpected error when opening stream: {}", e)
    }
}
Ok(conn.open_stream().is_err())
```

PR #42, opened 2019-02-16
Test half-closed streams.

romanb:
Test the behaviour of half-closed streams:

  * The side that initiates the shutdown can still read.
  * The side that receives the shutdown can still write.

Fails before https://github.com/paritytech/yamux/pull/38 and passes afterwards.

### DIFF
--- a/tests/smoke.rs
+++ b/tests/smoke.rs
@@ -9,20 +9,22 @@
 // at https://opensource.org/licenses/MIT.
 
 use bytes::Bytes;
-use futures::{future::{self, Either, Loop}, prelude::*};
+use futures::{future::{self, Either, Loop}, prelude::*, try_ready};
 use log::{debug, error};
 use quickcheck::{quickcheck, Arbitrary, Gen, QuickCheck, TestResult};
 use rand::Rng;
 use std::{fmt::{Debug, Display}, io, net::{Ipv4Addr, SocketAddr, SocketAddrV4}};
+use std::io::Write;
 use tokio::{
     codec::{BytesCodec, Framed, Encoder, Decoder},
     net::{TcpListener, TcpStream},
     runtime::Runtime
 };
-use yamux::{Config, Connection, Mode};
+use tokio_io::{AsyncWrite, AsyncRead};
+use yamux::{Config, Connection, ConnectionError, Mode, StreamHandle, State};
 
 #[test]
-fn prop_send_receive() {
+fn prop_send_recv() {
     fn prop(msgs: Vec<Msg>) -> TestResult {
         if msgs.is_empty() {
             return TestResult::discard()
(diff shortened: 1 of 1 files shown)


tests/smoke.rs
```diff
@@ -73,6 +71,49 @@ fn prop_max_streams() {
...
+        let cfg = Config::default();
+
+        // Server should be able to write on a stream shutdown by the client.
+        let server = server(cfg.clone(), l).and_then(|c| c
+            .into_future()
+            .map_err(|(e,_)| e)
+            .and_then(|(stream, _)| {
+                let mut s = stream.expect("S: No incoming stream");
+                future::poll_fn(move || {
+                    let mut buf = vec![0;1];
```
twittner:
```suggestion
                    let mut buf = [0;1];
```

tests/smoke.rs
```diff
@@ -73,6 +71,49 @@ fn prop_max_streams() {
...
+                    try_ready!(s.poll_read(&mut buf));
+                    assert!(s.state() == Some(State::RecvClosed));
+                    let w = s.write(&buf);
```
twittner: This and the next line are synchronous calls and may fail with `std::io::ErrorKind::WouldBlock`.

tests/smoke.rs
```diff
@@ -73,6 +71,49 @@ fn prop_max_streams() {
...
+                    s.flush()?;
+                    Ok(Async::Ready(w.is_ok()))
+                })
+            }).map_err(|e| error!("S: connection error: {}", e)));
+
+        // Client should be able to read after shutting down the stream.
+        let client = client(cfg, a).and_then(move |c| {
+            let mut s = new_stream(&c);
+            s.write(&vec![msg]).expect("C: Write failed");
```
twittner: Another blocking call.

tests/smoke.rs
```diff
@@ -73,6 +71,49 @@ fn prop_max_streams() {
...
+            future::poll_fn(move || {
+                try_ready!(s.shutdown());
+                assert!(s.state() == Some(State::SendClosed));
+                let mut buf = vec![0;1];
```
twittner:
```suggestion
                let mut buf = [0;1];
```

tests/smoke.rs
```diff
@@ -73,6 +71,49 @@ fn prop_max_streams() {
...
+                try_ready!(s.poll_read(&mut buf));
+                assert!(s.state() == Some(State::Closed));
+                Ok(Async::Ready(buf == vec![msg]))
```
twittner:
```suggestion
                Ok(Async::Ready(buf == [msg]))
```

tests/smoke.rs
romanb: I am aware, but only if the stream has no credit, which I considered here as a genuine test failure and hence thought it is fine to write synchronously as it simplifies the test. Do you disagree, i.e. would want the test to write asynchronously?

twittner: > I am aware, but only if the stream has no credit, which I considered here as a genuine test failure and hence thought it is fine to write synchronously as it simplifies the test.

If [flush](https://github.com/paritytech/yamux/blob/master/src/connection.rs#L716)  returns `WouldBlock` it has other reasons.

> Do you disagree, i.e. would want the test to write asynchronously?

Yes please. The test should only assume that the stream is non-blocking. The fact that the current implementation uses internal buffering and therefore only exhibits `WouldBlock` errors in a limited set of cases can change any time and is not something to rely on.

romanb: I changed the test to be based on futures via the `tokio::io` functions instead of using `poll_fn`. Apart from no longer using blocking IO, the test should look cleaner now and uses not just single byte messages but test `Msg`s of random length.

PR #44, opened 2019-03-01
Replace status boolean with `ConnStatus` enum.

twittner:
Also remove `Connection::shutdown` and fold its behaviour into `Connection::close`, i.e. the close will append a graceful `GoAway` frame to the buffer before proceeding to close the connection.

The new `ConnStatus` enum has three states, `Open`, `Shutdown` and `Closed`. Most places continue to distinguish only between `Open` and not open. `Shutdown` is set by `close` in case it can not complete the close but is notified again. Besides calling `close` again, only flushing the connection is possible in `Shtudown` mode.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -98,6 +98,11 @@ where
         Ok(Some(StreamHandle::new(id, buffer, self.clone())))
     }
 
+    #[deprecated(since = "0.1.8", note = "Use `Connection::close`.")]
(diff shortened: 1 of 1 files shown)


tomaka:
I suggest keeping a dummy `shutdown()` function marked as `#[deprecated]` and that just calls `close()`.
This way, it's not a breaking change and we don't have to publish a new major version.

PR #49, opened 2019-03-20
Add another test.

twittner:
This creates a lot of streams over a single connection, each sending and receiving one data item. Note that this test does sometimes not terminate which indicates a bug in this crate.


twittner:
Case in point: https://travis-ci.org/paritytech/yamux/jobs/508936700:

> test concurrent_streams ... test concurrent_streams has been running for over 60 seconds

PR #50, opened 2019-03-20
Rework implementation.

twittner:
Try a different implementation approach. Also fixes the sometimes not terminating test in `concurrent.rs` (cf. #49). In terms of performance the test in `concurrent.rs` now takes 1.8s vs. 6.6s on my machine (real time, 1000 streams each sending and receiving 1000 KiB of data, compiled in release mode).


src/connection.rs
```diff
@@ -338,427 +387,495 @@ where
...
+                                        });
+                                    return Box::new(future)
+                                }
+
+                                let window = entry.get().1.decrement_window(frame.body().len() as u32);
+                                entry.get().1.add_data(frame.into_body());
+
+                                // If the stream window is closed, reset the window and grant the
+                                // remote more credit so it can keep sending data.
+                                if window == 0 && admin.config.window_update_mode == WindowUpdateMode::OnReceive {
```
romanb: If the state of the stream is now `Closed`, which it may be if the stream was `ReadOnly` and this frame is a `FIN` (i.e. `is_finish` is true), then updating the window and sending a window update should probably  not be done?

twittner:
Thank you for your review!

> A `Handle` representing a connection does not appear to be cloneable, as compared to the previous `Connection`. [...]

Right, so I did not think of it as overly important. Users can always put a `Handle` into an `Arc<Mutex<_>>` if they want and not baking this into `Handle` directly makes the implementation simpler.

> [...] dropping a connection `Handle` seems to initiate a "graceful" `close()` of the connection, whereas I would have expected that to be a `reset()` [...]

We can change the `Drop` impl to trigger a connection abort. I agree that it would probably fit better here and users should call `close` if they wanted pending data to be sent to the remote before closing.

> Furthermore, `close()` seems to put all streams into `State::Closed` right away, whereas I would expect a graceful closing of a connection to do a half-close on all streams.

The idea behind `close` is to allow pending data to be sent before closing the connection as it is already committed from the perspective of the current streams. It is not a half-close over all streams as the connection has not enough information to make the decision to half-close a stream on behalf of the stream itself. Additionally, users who call `close` expect the connection to be closed soon-ish. Half-open streams may still produce incoming traffic for an indetermined amount of time. 

> A `FIN` is sent, the in-memory stream of outgoing frames ends, causing `Sender` to notify the `Connection` with `EndOfStream` which in turn causes the connection to send an `RST`, removing the stream from the connection. I.e. a `FIN` for a graceful close is immediately followed by an `RST`.

It is not necessarily the case that a half-close is *immediately* followed by a `Drop` of the in-memory stream because the yamux stream may very well be receiving incoming data for a while before being dropped. But we have no choice but to send a `RST` frame if the stream is not closed yet if we want to keep the remote informed that the stream is gone.

> The remote will reply to the `FIN` with another `FIN` but since the other end already removed the stream, that triggers "data for unknown stream" followed by another `RST` being sent.

We should change that to not send a `RST` for unknown streams if the incoming frame was a `FIN`.

> I think `fn end_of_stream` in `connection.rs` should only remove the stream and send a reset if it is in state `Open` or `WriteOnly`, whereas if it is in `ReadOnly` that indicates a graceful shutdown (half-close) and should keep the stream in that state until the `FIN` from the remote is received.

If (for whatever reason) we never receive the `FIN` from remote we would leak memory as we would never look at the stream again.

romanb:
> > A `Handle` representing a connection does not appear to be cloneable, as compared to the previous `Connection`. [...]
> 
> Right, so I did not think of it as overly important. Users can always put a `Handle` into an `Arc<Mutex<_>>` if they want and not baking this into `Handle` directly makes the implementation simpler.

I see. Sounds good to me!
 
> > Furthermore, `close()` seems to put all streams into `State::Closed` right away, whereas I would expect a graceful closing of a connection to do a half-close on all streams.
> 
> The idea behind `close` is to allow pending data to be sent before closing the connection as it is already committed from the perspective of the current streams. It is not a half-close over all streams as the connection has not enough information to make the decision to half-close a stream on behalf of the stream itself. Additionally, users who call `close` expect the connection to be closed soon-ish. Half-open streams may still produce incoming traffic for an indetermined amount of time.

I see. Contrary to what I said earlier, I think this is actually also the same as the old behaviour, since the previous connection `close()` does a graceful close on the TCP connection after flushing pending frames, but that will obviously not care about substreams, which are  not closed gracefully in the sense of a half-close. So this now makes sense to me, thanks for clarifying!

> > A `FIN` is sent, the in-memory stream of outgoing frames ends, causing `Sender` to notify the `Connection` with `EndOfStream` which in turn causes the connection to send an `RST`, removing the stream from the connection. I.e. a `FIN` for a graceful close is immediately followed by an `RST`.
> 
> It is not necessarily the case that a half-close is _immediately_ followed by a `Drop` of the in-memory stream because the yamux stream may very well be receiving incoming data for a while before being dropped. But we have no choice but to send a `RST` frame if the stream is not closed yet if we want to keep the remote informed that the stream is gone.

Ah, I didn't realise that the `RST` was sent because the stream got dropped immediately after the successful close in my test! Now it makes sense.

> > The remote will reply to the `FIN` with another `FIN` but since the other end already removed the stream, that triggers "data for unknown stream" followed by another `RST` being sent.
> 
> We should change that to not send a `RST` for unknown streams if the incoming frame was a `FIN`.

Sounds good.

twittner:
> > [...] dropping a connection `Handle` seems to initiate a "graceful" `close()` of the connection, whereas I would have expected that to be a `reset()` [...]
> 
> We can change the `Drop` impl to trigger a connection abort.

I was thinking more about it and the change to send `Abort` in `Drop` is wrong I think. A `Handle::close` followed by a `Handle::drop` will send messages `Close`, followed by `Abort` to the connection actor and in most cases this will be almost the same as sending `Abort` directly since the actor will not have finished the closing of the connection when being told to abort. Instead `Handle::abort` should be viewed as an alternative to `Handle::close` which can be invoked by clients if they wish to terminate the connection immediately, but usually `Handle::close` will be used (either explicitly or implicitly by `Drop`). If you (@romanb, @tomaka) agree, I would like to revert 64b9caaa0a09fd16622e6ab8c0bf7a515b683518.

romanb:
Sounds good to me.

twittner:
Here is a benchmark result of `master` relative to an optimised variant of this PR which uses the existential type feature to avoid some boxing:

Screenshot

So, while this PR avoids lock contention, the results show a significant overhead in all other cases. 
Therefore I am putting it on hold.

twittner:
Closing in favour of #59.

PR #51, opened 2019-03-26
Limit pending buffer and fix test.

twittner:
The cause for the sometimes not terminating test was that both endpoints
attempt to finish a writing operation and would not continue reading
from the socket. We must--as much as possible--continue reading, even
when we can not complete a write operation to prevent deadlock.

Since we buffer pending frames, which are meant to be delivered to the
remote and we make progress even if we can not write, we must limit the
resource growth of the pending frames buffer. This PR adds a new setting
`max_pending_frames` to do just that.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -56,6 +56,9 @@ quick_error! {
         TooManyStreams {
             display("maximum number of streams exhausted")
         }
+        TooManyPendingFrames {
+            display("maximum number of pending streams reached")
+        }
         #[doc(hidden)]
         __Nonexhaustive
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -57,7 +57,7 @@ quick_error! {
             display("maximum number of streams exhausted")
         }
         TooManyPendingFrames {
-            display("maximum number of pending streams reached")
+            display("maximum number of pending frames reached")
         }
         #[doc(hidden)]
         __Nonexhaustive
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,9 +111,8 @@ impl Config {
         self.max_num_streams = n
     }
 
-    /// Set the max. number of pending frames, i.e. frames
-    /// which are to be sent to the remote, but can not right
-    /// away and need to be buffered.
+    /// Set the max. number of pending frames, i.e. outgoing
+    /// frames which have not yet been sent.
     pub fn set_max_pending_frames(&mut self, n: usize) {
         self.max_pending_frames = n
     }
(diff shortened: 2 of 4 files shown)


src/error.rs
```diff
@@ -56,6 +56,9 @@ quick_error! {
         TooManyStreams {
             display("maximum number of streams exhausted")
         }
+        TooManyPendingFrames {
+            display("maximum number of pending streams reached")
```
romanb: Shouldn't that read "Maximum number of pending __frames__ reached"?

src/lib.rs
```diff
@@ -82,6 +84,7 @@ impl Default for Config {
             receive_window: DEFAULT_CREDIT,
             max_buffer_size: 1024 * 1024,
             max_num_streams: 8192,
+            max_pending_frames: 2048,
```
romanb: It is probably difficult to come up with criteria for a good default value, but how about aligning it with `max_num_streams` by the simple reasoning that there should be on average at least 1 free slot for an outgoing frame per stream, i.e. the max number of streams loosely defines the "bandwidth" in terms of the number of frames sent at a time for a healthy connection?

src/lib.rs
```diff
@@ -108,6 +111,14 @@ impl Config {
         self.max_num_streams = n
     }
 
+    /// Set the max. number of pending frames, i.e. frames
+    /// which are to be sent to the remote, but can not right
+    /// away and need to be buffered.
```
romanb: Is it not the case that all frames to be sent go through the buffer, i.e. the `pending` queue? The comment may then be slightly misleading as one may think that even having a 0-sized buffer would be a valid configuration, only causing errors when a frame cannot be sent "right away" and needs to be buffered. It may then be better to just say something like "[..], i.e. outgoing frames which have not yet been sent."? If a `0` value for this option is not sensible, then using `NonZeroUsize` at least for the parameter may also be worth considering.

src/lib.rs
twittner: It is indeed not easy to find a good default. The thing to keep in mind is the upper bound of memory usage we want to allow.  This would be `max_pending_frames * receive_window` (of the remote and ignoring 12 bytes for the header per frame). Using the proposed default settings this would amount to 512 MiB. With `max_pending_frames = 8192` it would go up to 2 GiB. I have no strong opinion which one to chose, but considering that processes may use more than one connection this could easily make them prime targets for an OOM killer. But maybe 512 MiB per connection would already be too much. I do not know.

PR #58, opened 2019-10-22
Format code using 'cargo fmt'

Atul9:


Atul9:
[clabot:check]

Atul9:
Hi @twittner can you please share why was this PR closed?

twittner:
I rejected this PR because the current format is actually quite thoughtful, so I can not approve bulldozing it with some tool that does not understand any of it.

PR #59, opened 2019-11-01
Upgrade to async/await and futures-preview.

twittner:
This PR reworks the implementation to use async/await and futures-preview (0.3.0-alpha.19). It supersedes #50, but represents an equally big change, so instead of reading the diff it is probably easier to just read the whole source code. Compared to the current `master` version the following changes are noteworthy:

- `Config::max_pending_frames` is gone. Since internally an MPSC channel is used to communicate outgoing messages to the `Connection`, which is ultimately responsible for delivering them, we use an internal upper bound which (due to the way `futures` MPSC channels behave) is rather soft, i.e. it increases by one with every `Stream`. When reached, we do not error, but back pressure is exercised and `Stream`s wait until there is capacity.
- It is no longer the case that every `Stream` tries to acquire the `Mutex` around the `Connection` in order to make progress (In fact, `Connection` is no longer put into a `Mutex`. The contention this created is presumably one of the reasons why performance suffers when many streams are in use). Only each `Stream` contains a `Mutex` around its shared state that is updated by the `Connection` when necessary. Only the connection drives the underlying I/O resource so it is necessary to keep calling `Connection::next_stream` in order to make progress.
- We no longer always reset a `Stream` when it is dropped. Instead we analyse the `Stream` state to figure out how to gracefully terminate the stream (see `Connection::garbage_collect` for details).
- We no longer hold a single collection with all tasks that want to be notified. Instead each `Stream` now stores a `Waker` for reading and one for writing if it can not make progress. Whereas we previously notified all tasks when new data arrived, we now wake only the `Waker`s of `Streams` which can actually make progress (e.g. read data because more has arrived).

There are probably more changes which could be mentioned but these are the ones from the top of my head for now. The connection module contains an [overview](https://github.com/twittner/yamux/blob/develop/src/connection.rs#L11-L51) about the general implemenation approach and some potential improvements that could be made in the future.

FWIW, here are the benchmark results against `master`. Since a lot has changed in tokio, the numbers are not really comparable, but at least they show that we are not slower than we used to be:

bench


twittner:
@romanb: Contrary to what I thought, this PR is not yet ready for review but should still be considered work in progress.

twittner:
@romanb: Should be good now (see 560dc67858521aac3e761e498d8e150846135c1d and a2eeb7348d28c2fc03d12c958e27b1c2605ac5ca).

src/connection.rs
```diff
@@ -8,793 +8,841 @@
...
+                    Either::Right(self.control_receiver.next())
+                };
+
+            if num_terminated == 3 {
+                log::error!("{}: all futures are terminated", self.id);
+                return Err(ConnectionError::Closed)
+            }
+
+            let next_item =
+                future::poll_fn(move |cx: &mut Context| {
```
tomaka: There's a `maybe_done` wrapper that could probably do the same.
Something like that:

```
let a = future::maybe_done(next_stream_command);
let b = future::maybe_done(next_control_command);
let c = future::maybe_done(next_inbound_frame);
future::select(future::select(a, b), c).await;
if let MaybeDone::Done(a) = a {
    ...
}
// same for b and c
```

However your code might actually be more readable, so I'm not sure.
In general, selecting over multiple futures is still something that hasn't really figured out in the Rust async ecosystem yet, I think.

src/connection.rs
```diff
@@ -8,793 +8,841 @@
...
+                    if a.is_pending() && b.is_pending() && c.is_pending() {
+                        return Poll::Pending
+                    }
+                    Poll::Ready((a, b, c))
+                });
+
+            let (stream_command, control_command, inbound_frame) = next_item.await;
+
+            if let Poll::Ready(cmd) = control_command {
+                self.on_control_command(cmd).await?
```
tomaka: This is a bit dangerous. If the future is interrupted (i.e. the user polls the future returned by `next()` then destroys that future before it ends), then we are losing the `stream_command` and `inbound_frame` (and might end up with an inconsistent state?).

I might be a better idea to store the pending commands in the `Connection` across yield points, rather than local variables.

twittner: As discussed in person, we will not support cancellation of `Connection::next_stream` for the time being. A  comment [was added](https://github.com/paritytech/yamux/pull/59/commits/3dd2f969073f0ecd649f09a68e507a51e8cbe895) to make this explicit.

romanb: While it would not be surprising to me that things may break if futures returned from `next_stream` are not polled to completion given that `next_stream` is the overall driving force of the connection, it would just be good to clearly document that fact.

src/connection/stream.rs
```diff
@@ -0,0 +1,308 @@
...
+            (Open,                _) => self.state = next,
+            (RecvClosed,     Closed) => self.state = Closed,
+            (RecvClosed,       Open) => {}
+            (RecvClosed, RecvClosed) => {}
+            (RecvClosed, SendClosed) => self.state = Closed,
+            (SendClosed,     Closed) => self.state = Closed,
+            (SendClosed,       Open) => {}
+            (SendClosed, RecvClosed) => self.state = Closed,
+            (SendClosed, SendClosed) => {}
+        }
```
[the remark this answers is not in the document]
twittner: Yes, I actually prefer less catch-alls here.

src/connection.rs
```diff
@@ -8,793 +8,841 @@
...
+#[derive(Debug)]
+enum ClosingState {
+    /// We are open for business.
+    None,
+    /// We have received a `ControlCommand::Close` and are shutting
+    /// down operations. The `Sender` will be informed once we are done.
+    Started(oneshot::Sender<()>),
+    /// The shutdown is complete and we are closed for good.
+    Complete
 }
```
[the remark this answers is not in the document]
twittner: I changed it to `Shutdown { NotStarted, InProgress(..), Complete }` because `State` is already used for the `Stream` state. I hope you approve.

src/connection.rs
```diff
@@ -8,793 +8,857 @@
...
-        if self.pending.len() >= self.config.max_pending_frames {
-            return Err(ConnectionError::TooManyPendingFrames)
+    /// Get the next inbound `Stream` and make progress along the way.
+    ///
+    /// This is called from `Connection::next_stream` instead of being a
+    /// public method itself in order to guarantee proper closing in
+    /// case of an error or at EOF.
+    async fn next(&mut self) -> Result<Option<Stream>> {
+        loop {
+            self.garbage_collect().await?;
```
romanb: Can this not happen outside of the loop? Can the `strong_count` of a stream go down within the loop (I didn't see how)?

romanb: Ah, I guess streams can be dropped between yield points (i.e. `.await`).

twittner: The strong count of a `Stream`'s `Arc` can change at any time by just dropping the `Stream`. If we were to garbage collect only once outside the loop, we risk adding many new (outbound) streams (depending on the number of control commands) while not removing dropped ones.

PR #65, opened 2019-11-27
Update to bytes-0.5

twittner:
Some notable changes:

- `BytesMut::bytes_mut` now returns a `&mut [MaybeUninit<u8>]` as it may contain uninitialised memory. Since we resize our buffer, which implies initialisation, we can just cast this slice to a `&mut [u8]` and use it directly.

- `BytesMut::remaining_mut` no longer gives the length of the `BytesMut::bytes_mut` as the `BufMut` implementation grows the buffer as needed. So, to figure out if we need to reserve another BLOCKSIZE chunk we now check how much capacity is left.

- `futures_codec` has been removed from tests as it has not updated to bytes-0.5. The tests now work directly with the stream as `AsyncRead` and `AsyncWrite` impl.

### DIFF
--- a/src/frame/io.rs
+++ b/src/frame/io.rs
@@ -11,7 +11,7 @@
 use bytes::{BufMut, BytesMut};
 use crate::u32_as_usize;
 use futures::{io::BufWriter, prelude::*, ready};
-use std::{io, pin::Pin, task::{Context, Poll}};
+use std::{io, mem::MaybeUninit, pin::Pin, task::{Context, Poll}};
 use super::{Frame, header::{self, HeaderDecodeError}};
 use thiserror::Error;
 
@@ -93,7 +93,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Stream for Io<T> {
             match self.decode() {
                 Ok(Some(f)) => return Poll::Ready(Some(Ok(f))),
                 Ok(None) => {
-                    if self.buffer.remaining_mut() < BLOCKSIZE {
+                    if self.buffer.capacity() - self.buffer.len() < BLOCKSIZE {
                         let n = self.buffer.len();
                         self.buffer.resize(n + BLOCKSIZE, 0);
                         unsafe { self.buffer.set_len(n) }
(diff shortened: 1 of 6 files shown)


src/frame/io.rs
```diff
@@ -93,14 +93,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Stream for Io<T> {
             match self.decode() {
                 Ok(Some(f)) => return Poll::Ready(Some(Ok(f))),
                 Ok(None) => {
-                    if self.buffer.remaining_mut() < BLOCKSIZE {
+                    if self.buffer.capacity() - self.buffer.len() < BLOCKSIZE {
                         let n = self.buffer.len();
                         self.buffer.resize(n + BLOCKSIZE, 0);
                         unsafe { self.buffer.set_len(n) }
                     }
                     unsafe {
```
tomaka: Why do we have unsafe here? I'd take a performance hit any day over unsafe code.

src/frame/io.rs
```diff
@@ -93,14 +93,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Stream for Io<T> {
             match self.decode() {
                 Ok(Some(f)) => return Poll::Ready(Some(Ok(f))),
                 Ok(None) => {
-                    if self.buffer.remaining_mut() < BLOCKSIZE {
+                    if self.buffer.capacity() - self.buffer.len() < BLOCKSIZE {
                         let n = self.buffer.len();
                         self.buffer.resize(n + BLOCKSIZE, 0);
                         unsafe { self.buffer.set_len(n) }
```
tomaka: What about `truncate`, which also happens to be safe?

src/frame/io.rs
twittner: We need `unsafe` here because of `bytes_mut()` and `advance_mut()`. What we want is reading data into the buffer into the range [len(), capacity()). I do not know an API in `bytes` which would allow us to do this without `unsafe`.

PR #73, opened 2020-01-29
Send `ACK`s and enable deferred `SYN`s.

twittner:
This PR changes two aspects related to newly opened inbound and outbound streams:

1. Up to now we never replied to opened inbound streams with an `ACK` flag, despite the fact that the spec clearly demands this behaviour. While we continue to ignore any `ACK` flags we receive, with this change we properly set an `ACK` flag if required.
2. The current behaviour is to always immediately send a `WindowUpdate` frame to the remote when opening an outbound stream. While we do not wait for an acknowledgement before we continue to send data over the stream, we could also hold off and set the initial `SYN` flag in the first data frame that is being sent over the new stream to the remote. The newly added option `lazy_open` enables this. This should eventually become the new default.

Additionally, this PR makes casting of header and frame types more principled. Instead of arbitrary casts only casts from `()` to `Data`, `WindowUpdate` and `Ping` are allowed. We also model a binary sum type and provide introductions from `T` to `Either<T, U>` and `T` to `Either<U, T>` but no projections. This allows a more type-safe `StreamCommand` whose `SendFrame` constructor only accepts `Frame<Either<Data, WindowUpdate>>` values which is what we expect coming from streams.

### DIFF
--- a/src/frame/header.rs
+++ b/src/frame/header.rs
@@ -67,12 +67,12 @@ impl<T> Header<T> {
         }
     }
 
-    /// Introduce this header to the left of a binary header type.
+    /// Introduce this header to the right of a binary header type.
     pub(crate) fn right<U>(self) -> Header<Either<U, T>> {
         self.cast()
     }
 
-    /// Introduce this header to the right of a binary header type.
+    /// Introduce this header to the left of a binary header type.
     pub(crate) fn left<U>(self) -> Header<Either<T, U>> {
         self.cast()
     }
@@ -438,4 +438,3 @@ mod tests {
             .quickcheck(property as fn(Header<()>) -> bool)
     }
 }
-
--- a/src/frame/header.rs
+++ b/src/frame/header.rs
@@ -79,17 +79,17 @@ impl<T> Header<T> {
 }
 
 impl Header<()> {
-    pub(crate) fn as_data(self) -> Header<Data> {
+    pub(crate) fn into_data(self) -> Header<Data> {
         debug_assert_eq!(self.tag, Tag::Data);
         self.cast()
     }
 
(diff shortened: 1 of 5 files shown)


twittner:
> I admit that the logic of this crate is a bit over my head

I tried to explain the structure and implementation of the crate in various comments, e.g. [here](https://github.com/paritytech/yamux/blob/ab12b4717db516f8d1366961a1f2d5a9d948e98e/src/lib.rs#L11-L23) or [here](https://github.com/paritytech/yamux/blob/ab12b4717db516f8d1366961a1f2d5a9d948e98e/src/connection.rs#L11-L88) and to smaller extends in many other places. Please let me know if there is anything unclear or could be improved.

tomaka:
Sorry, that's not a critic of the crate. It's just that yamux is totally out of my head right now and that I would have to spend hours studying how things work.

src/frame/header.rs
```diff
@@ -64,6 +66,33 @@ impl<T> Header<T> {
             _marker: std::marker::PhantomData
         }
     }
+
+    /// Introduce this header to the left of a binary header type.
```
romanb:
```suggestion
    /// Introduce this header to the right of a binary header type.
```

src/frame/header.rs
```diff
@@ -64,6 +66,33 @@ impl<T> Header<T> {
...
+    pub(crate) fn right<U>(self) -> Header<Either<U, T>> {
+        self.cast()
+    }
+
+    /// Introduce this header to the right of a binary header type.
```
romanb:
```suggestion
    /// Introduce this header to the left of a binary header type.
```

PR #79, opened 2020-03-20
Add check for BrokenPipe io error

sdbondi:
If there is a broken pipe from the socket, the connection should close.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -590,6 +590,10 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 log::debug!("{}: socket eof", self.id);
                 Err(ConnectionError::Closed)
             }
+            Err(e) if e.io_kind() == Some(std::io::ErrorKind::BrokenPipe) => {
+                log::debug!("{}: broken pipe", self.id);
+                Err(ConnectionError::Closed)
+            }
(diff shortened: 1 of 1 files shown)


sdbondi:
[clabot:check]

twittner:
> If there is a broken pipe from the socket, the connection should close.

The connection does already close if there is a broken pipe error. In fact the connection closes for any kind of I/O error caused by the socket (cf. [[1]], [[2]]).

[1]: https://github.com/paritytech/yamux/blob/def2093c9a8b11d482845ccfb9dd40689496a0ad/src/connection.rs#L310-L316
[2]: https://github.com/paritytech/yamux/blob/def2093c9a8b11d482845ccfb9dd40689496a0ad/src/connection.rs#L426

twittner:
I have opened https://github.com/paritytech/yamux/pull/80 to return early with `Ok(None)` from `Connection::next_stream` if the methods is repeatedly invoked after a previous `Err(_)` or `Ok(None)`. Would that be helpful?

sdbondi:
Ok thanks - I did find an edge case I'd missed in my code which caused next stream to be called after an error, which I noticed and fixed after submitting this PR. `Connection::next_stream` returning `Ok(None)` was what I expected to happen if the socket closed and would have prevented a spinning loop - so I vote that it goes in :) Thanks again!

PR #81, opened 2020-06-11
Fix race on final FIN or RST frames.

romanb:
When a stream is dropped after sending some data, the corresponding `SendFrame` commands may still be in the
queue when stream garbage collection kicks in. Stream garbage collection conditionally sends final FIN or RST frames, but it does so directly on the socket, thus these frames may be sent (and hence received) before other frames that may have been sent before the stream was dropped, giving the remote a premature EOF.

The proposal here to address this problem is to send the final FIN or RST frames through the `stream_sender`, so they are guaranteed to be sent last, since the stream has already been dropped. As a side-effect, these frames being sent also turn
up in the trace logs, which was not the case before and made it a bit more difficult to diagnose.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ static_assertions = "1"
 
 [dev-dependencies]
 anyhow = "1"
-async-std = "1"
+async-std = "~1.5"
 criterion = "0.3"
 futures = "0.3.4"
 quickcheck = "0.9"
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -830,7 +830,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     State::Open => {
                         let mut header = Header::data(stream_id, 0);
                         header.rst();
-                        Some(Frame::new(header))
+                        Some(Frame::new(header).left())
                     }
                     // The stream was dropped without calling `poll_close`.
                     // We have already received a FIN from remote and send one
@@ -838,7 +838,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     State::RecvClosed => {
                         let mut header = Header::data(stream_id, 0);
                         header.fin();
-                        Some(Frame::new(header))
+                        Some(Frame::new(header).left())
                     }
                     // The stream was properly closed. We either already have
                     // or will at some later point send our FIN frame.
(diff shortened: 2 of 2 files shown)


romanb:
I am wondering, since the `stream_receiver` can be closed before the connection has completed the full "shutdown", whether the following may be preferable at the end of `garbage_collect()`:
```rust
            if let Some(f) = frame {
                let cmd = StreamCommand::SendFrame(f);
                if let Err(e) = self.stream_sender.try_send(cmd) {
                    if let StreamCommand::SendFrame(f) = e.into_inner() {
                        self.socket.get_mut().send(&f).await.or(Err(ConnectionError::Closed))?
                    }
                }
            }
```
Since I think if the `stream_receiver` has been dropped, it is assured that all stream commands have been processed already and it is safe to try to send the final frame through the socket, which may still be open if I'm not mistaken.

romanb:
Falling back to using the `socket` if it is still open may be trickier than I thought, since it should only happen if the `Sender` is disconnected, not if the channel is full. I could currently only come up with:
```rust
            if let Some(f) = frame {
                let f_alt = f.clone();
                let cmd = StreamCommand::SendFrame(f);
                if let Err(e) = self.stream_sender.send(cmd).await {
                    debug_assert!(e.is_disconnected());
                    self.socket.get_mut().send(&f_alt).await.or(Err(ConnectionError::Closed))?
                }
            }
```

twittner:
> When a stream is dropped after sending some data, the corresponding `SendFrame` commands may still be in the
> queue when stream garbage collection kicks in. Stream garbage collection conditionally sends final FIN or RST frames, but it does so directly on the socket, thus these frames may be sent (and hence received) before other frames that may have been sent before the stream was dropped, giving the remote a premature EOF.

A RST or FIN frame can only be sent prior to an enqueued data frame of the stream if the stream was not closed properly. (A point of debate may be if we should really send a RST when the stream was properly closed but there is lingering data in the stream buffer.)

> The proposal here to address this problem is to send the final FIN or RST frames through the `stream_sender`, so they are guaranteed to be sent last, since the stream has already been dropped.

We can not add to the `stream_sender` as this could result in a deadlock if the channel if full.

romanb:
> We can not add to the `stream_sender` as this could result in a deadlock if the channel if full.

You're right, I just noticed that as well.

romanb:
> A RST or FIN frame can only be sent prior to an enqueued data frame of the stream if the stream was not closed properly. 

Good point, so closing a substream gracefully instead of just dropping it is really obligatory. The context in which this appeared is the ping protocol in libp2p where [a substream is not properly closed](https://github.com/libp2p/rust-libp2p/blob/9aaf04241008a7de9007c644ffec15a3bacbb24b/protocols/ping/src/protocol.rs#L70). I will open a PR there and subsequently this PR can probably be closed.

PR #82, opened 2020-06-11
Fix read-after-close?

romanb:
Currently the `Stream` and `AsyncRead` impls for streams check the condition for read-after-close based on whether
the stream command sender is closed, but that never seems to be the case, unless the receiver is dropped which only happens when the connection closes. I'm not sure whether this is intentional. Should read-after-close on a substream always work as long as the connection is not closed?

As a preliminary suggestion, I've changed the condition to check the substream state instead of whether the sender is closed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ static_assertions = "1"
 
 [dev-dependencies]
 anyhow = "1"
-async-std = "1"
+async-std = "~1.5"
 criterion = "0.3"
 futures = "0.3.4"
 quickcheck = "0.9"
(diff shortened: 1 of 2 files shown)


romanb:
Note that I've updated the PR description as this may in fact be a non-issue / a misunderstanding of the scope of read-after-close.

romanb:
From the documentation on `set_read_after_close`:
```rust
    /// Allow or disallow streams to read from buffered data after
    /// the connection has been closed.
```
so this is probably deliberately only related to closing the connection, not a substream, so I'm going to proactively close this.

PR #83, opened 2020-06-12
Use tokio in tests.

twittner:


mxinden:
Out of curiosity, could you expand on the motivation for this patch?

twittner:
Like in https://github.com/libp2p/rust-libp2p/issues/1588, tests do not complete with async-std > 1.5.0.

src/frame/io.rs
```diff
@@ -232,7 +232,7 @@ mod tests {
     #[test]
     fn encode_decode_identity() {
         fn property(f: Frame<()>) -> bool {
-            async_std::task::block_on(async move {
+            futures::executor::block_on(async move {
```
mxinden, line 235: Why use `futures::executor` here and `Runtime::block_on` in other places?

twittner: The correctness does not depend on the runtime choice so it does not really matter. As this is a small property test without concurrency, executing it on the same thread is more light-weight.

twittner:
NB: CI [is failing](https://travis-ci.org/github/paritytech/yamux/jobs/697587466) with nightly because of https://github.com/rust-lang/cargo/issues/8351:

```
$ cargo build --verbose
    Updating crates.io index
error: failed to download `futures v0.3.5`
Caused by:
  unable to get packages from source
Caused by:
  failed to parse manifest at `/home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-0.3.5/Cargo.toml`
Caused by:
  readme file with name '../README.md' was not found
```

PR #87, opened 2020-07-02
Consume and return `self` in `next_stream`.

twittner:
This ensures that futures are not cancelled and that the previous call finished successfully.

This PR also returns `Stream` instead of `Option<Stream>` and signals EOF via `ConnectionError::Closed` instead of `None`. This is motivated by the fact that errors already mean the connection should be considered closed and it is now impossible even to continue calling `next_stream`.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -170,8 +170,7 @@ pub struct Connection<T> {
     stream_sender: mpsc::Sender<StreamCommand>,
     stream_receiver: mpsc::Receiver<StreamCommand>,
     garbage: Vec<StreamId>, // see `Connection::garbage_collect()`
-    shutdown: Shutdown,
-    is_closed: bool
+    shutdown: Shutdown
 }
 
 /// `Control` to `Connection` commands.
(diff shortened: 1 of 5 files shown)


src/connection.rs
```diff
@@ -354,19 +349,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             }
         }
 
-        if let Err(ConnectionError::Closed) = result {
-            return Ok(None)
-        }
-
-        result
+        result.map(move |stream| (stream, self))
```
romanb, line 352: `result` is always `Err` here right, so the `map` just aligns the types? Presumable you want to avoid increasing the indentation of all the above code up to `self.is_closed = true` if it were put in an `Err(e) => ` branch of a `match` on the `result`?

twittner: I am fine with using a `match` instead. 99ec2b0 reformats the code accordingly.

twittner:
I am not sure if the API is not too inconvenient to use, so I close this for the time being.

Issue #90, opened 2020-08-10
Yamux connection issue with go


AgeManning:
I haven't fully tracked down the reason for this. But I'm seeing connection drops when using rust-libp2p with go-libp2p over yamux. 

Specifically we hit this line: https://github.com/paritytech/yamux/blob/develop/src/connection.rs#L673

With this error: 
```
3106e0cb/10: frame body larger than window of stream
```

I'll be chasing up the go-side also, just wondering if there was some extra insight on this?

twittner:
Thank you for your report! Could you provide logs or a test case that I could use to reproduce this?

AgeManning:
It happens intermittently in a complicated environment. We'll go digging a bit and report back with more info. 

Was just fishing for potentially a known issue or something obvious that could be wrong.

twittner:
@AgeManning, Did you manage to get some more information?

AgeManning:
We are still looking into this. There was a recent explosion of sorts on our testnet, so have been dealing with that. I temporarily disabled yamux to solve our problems, but we will start dedicating some time to debugging the yamux issue. 

Sorry for the delay, but will report back once we have more info/have found the source of the problem

pawanjay176:
This looks like an issue with the `receive_window` size. 
Even though the spec says to use 256 kb as the window size, go-libp2p-yamux by default has the param set to to 16mb because of performance issues
https://github.com/libp2p/go-libp2p-yamux/blob/master/transport.go#L14-L20

I might be wrong on this, but it looks to me that go-yamux doesn't adaptively adjust the receive window size for a stream based on how much data is coming in.

I was able to make rust work with go by simply by setting `config.receive_window` to `16 * 1024 * 1024`.

twittner:
Maybe you could provide logs on trace level which show the error you are getting? Without logs or some test case I am not sure how I could help you.

pawanjay176:
[Here](https://pastebin.com/pmsCnK9a) are the truncated trace logs for the stream that errored. 

It contains this additional debug log before this line https://github.com/paritytech/yamux/blob/develop/src/connection.rs#L689
```rust
log::debug!("Window size: {}, credit: {}", shared.window, shared.credit);
```

twittner:
[...]

PR #91, opened 2020-08-25
Send window update for initial data.

twittner:
PR #73 introduced a lazy open option to defer sending the initial frame when actual data is sent. If enabled and the initial data size equals DEFAULT_CREDIT (i.e. 256 KiB), it is necessary to immediately send a window update back when accepting such a stream, but only if window update mode is `OnReceive`, otherwise the update will be sent when the stream data is consumed.

With more options there are more ways how things can influence each other, so I have changed some tests to work with randomly generated configurations to test more combinations.


romanb (approved):
:+1: Does this relate to https://github.com/paritytech/yamux/issues/90?

twittner:
No. That issue is about go-libp2p-yamux overdrawing its credit.

romanb (approved):

PR #101, opened 2021-01-29
src/tests: Write and read concurrently

mxinden:
* src/tests: Write and read concurrently

   Most of the tests in `src/tests.rs` have the structure of a client
   sending data to a server which in turn echoes the data back to the
   client which reads it in full and compares it to what it initially sent.
   
   Some of the tests use randomly generated configurations where the
   `WindowUpdateMode` can be either `OnRead` or `OnReceive`. When using
   `WindowUpdateMode::OnRead` the client can not first send a whole message to
   the server and only then receive the whole message from the server. Say that
   a single message exceeds the window between server and client. In such case
   the server would eventually exhaust its window to the client, thus not being
   able to forward bytes to the client, thus not accepting new bytes from the
   client, thus the client would be blocked on sending and thus the whole test
   would deadlock.
   
   With this commit the client writes and reads concurrently and thus does
   not run into a deadlock when executing with `WindowUpdateMode::OnRead`.

* src/tests: Restrict number of quickcheck runs

   With quickcheck iterations potentially taking long, restrict number of
   iterations per test to 10.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -11,6 +11,7 @@
 use crate::{Config, Connection, ConnectionError, Mode, Control, connection::State};
 use crate::WindowUpdateMode;
 use futures::{future, prelude::*};
+use futures::io::AsyncReadExt;
 use quickcheck::{Arbitrary, Gen, QuickCheck, TestResult};
 use rand::Rng;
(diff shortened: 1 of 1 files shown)


romanb (approved):
Looks good to me :+1: .

PR #104, opened 2021-02-04
[tests] Add a test for a potential write deadlock.

romanb:
See https://github.com/paritytech/yamux/pull/102 for the preceding discussion. This PR adds a test that provokes a deadlock between two endpoints both wanting to write but cannot (i.e. `Pending`), though each could allow the other to make progress by reading. The test is disabled for now, due to actually stalling. It can be run with
```
RUST_LOG=yamux=debug cargo test deadlock -- --ignored
```
The logs will end with (`S` is the "server" echoing what the client sends, `C` is the "client" sending data it expects to get back, the "channel" is the artificially bounded "network connection")
```
[2021-02-04T15:13:35Z DEBUG yamux::tests::bounded] S: write: channel full, registering waker
...
[2021-02-04T15:13:35Z DEBUG yamux::tests::bounded] C: write: channel full, registering waker
```
i.e. both sides want to be woken to continue writing but neither actually does wake the other by reading, though the Client should do so because it reads and writes concurrently in separate tasks, but the Yamux connection insists on `.await`ing the write before doing anything else.

Unless someone sees a fault in the test, the next step would be to prepare a patch. As for the likelihood of this situation occurring, I'm not sure. It seems it would only occur if both endpoints get `Poll::Pending` trying to write a frame for some substream simultaneously while at the same time only being able to continue writing when the remote "makes room" on the connection by reading, hence the deadlock.

As an aside, this PR updates `tokio` in the tests.

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -209,6 +209,8 @@ fn prop_send_recv_half_closed() {
 /// This test simulates two endpoints of a Yamux connection which may be unable to
 /// write simultaneously but can make progress by reading. If both endpoints
 /// don't read in-between trying to finish their writes, a deadlock occurs.
+//
+// Ignored for now as the current implementation is prone to the deadlock tested below.
 #[test]
 #[ignore]
 fn write_deadlock() {
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -257,12 +257,8 @@ fn write_deadlock() {
         }).boxed().into()
     ).unwrap();
 
-    // Handles to tasks on which to wait for completion.
-    let mut handles = Vec::new();
-
     // Send the message, expecting it to be echo'd.
-    let msg = msg.clone();
-    handles.push(pool.spawner().spawn_with_handle(async move {
+    pool.run_until(pool.spawner().spawn_with_handle(async move {
         let stream = ctrl.open_stream().await.unwrap();
         let (mut reader, mut writer) = AsyncReadExt::split(stream);
         let mut b = vec![0; msg.len()];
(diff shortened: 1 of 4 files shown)


src/tests.rs
```diff
@@ -199,6 +206,85 @@ fn prop_send_recv_half_closed() {
...
+    pool.spawner().spawn_obj(
+        crate::into_stream(client).for_each(|_| {
+            panic!("Unexpected inbound stream for client");
+            #[allow(unreachable_code)]
+            future::ready(())
+        }).boxed().into()
+    ).unwrap();
+
+    // Handles to tasks on which to wait for completion.
+    let mut handles = Vec::new();
```
mxinden, line 261: Given that there is only ever a single handler, is the `Vec` and the `for` loop below necessary?

src/tests.rs
```diff
@@ -199,6 +206,85 @@ fn prop_send_recv_half_closed() {
     QuickCheck::new().tests(7).quickcheck(prop as fn(_))
 }
 
+/// This test simulates two endpoints of a Yamux connection which may be unable to
+/// write simultaneously but can make progress by reading. If both endpoints
+/// don't read in-between trying to finish their writes, a deadlock occurs.
```
mxinden, line 211:
```suggestion
/// don't read in-between trying to finish their writes, a deadlock occurs.
//
// Ignored for now as the current implementation is prone to the deadlock tested below.
```

With the above would not have to look at the git history.

mxinden (approved):
Thanks for the improved version.

src/tests.rs
romanb: That is a left-over :+1: .

romanb: Addressed in https://github.com/paritytech/yamux/pull/104/commits/2ed415012803c22d65d539bc5c7417cbda9dfd91.

PR #109, opened 2021-02-08
src/connection: Send WindowUpdate message early

mxinden:
To prevent senders from being blocked every time they have sent
RECEIVE_WINDOW (e.g. 256 KB) number of bytes, waiting for a WindowUpdate
message granting further sending credit, this commit makes Yamux send a
WindowUpdate message once half or more of the window has been received
(and consumed in `WindowUpdateMode::OnRead`). Benchmarking shows that
sending WindowUpdate messages early prevents senders from being blocked
waiting for additional credit on common network types.

For a detailed discussion as well as various benchmark results see
https://github.com/paritytech/yamux/issues/100.

Next to the above, this commit includes the following changes:

- Use `WindowUpdateMode::OnRead` in benchmarks. I would argue that
`OnRead` should be the default setting, given the importance of
flow-control. With that in mind, I suggest to benchmark that default
case.

- Ignore WindowUpdate messages for unknown streams. With this commit
WindowUpdate messages are sent more agressively. The following scenario
would surface: A sender would close its channel, eventually being
garbage collected. The receiver, before receiving the closing message,
sends out a WindowUpdate message. The sender receives the WindowUpdate
message not being able to associate it to a stream, thus resetting the
whole connection. This commit ignores WindowUpdate messages for unknown
streams instead.

When sending very large messages, WindowUpdate messages are still not
returned to the sender in time, thus the sender still being blocked in
such case. This can be prevented by splitting large messages into
smaller Yamux frames, see
https://github.com/paritytech/yamux/issues/100#issuecomment-774461550.
This additional optimization can be done in a future commit without
interfering with the optimization introduced in this commit.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -693,7 +693,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 shared.update_state(self.id, stream_id, State::RecvClosed);
             }
             let max_buffer_size = self.config.max_buffer_size;
-            if shared.buffer.len().map(move |n| n >= max_buffer_size).unwrap_or(true) {
+            if shared.buffer.len() >= max_buffer_size {
                 log::error!("{}/{}: buffer of stream grows beyond limit", self.id, stream_id);
                 let mut header = Header::data(stream_id, 0);
                 header.rst();
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -658,7 +658,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 let sender = self.stream_sender.clone();
                 Stream::new(stream_id, self.id, config, credit, credit, sender)
             };
-            let window_update;
+            let mut window_update = None;
             {
                 let mut shared = stream.shared();
                 if is_finish {
@@ -666,16 +666,12 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 }
                 shared.window = shared.window.saturating_sub(frame.body_len());
                 shared.buffer.push(frame.into_body());
-                if !is_finish
-                    && shared.window == 0
-                    && self.config.window_update_mode == WindowUpdateMode::OnReceive
-                {
-                    shared.window = self.config.receive_window;
-                    let mut frame = Frame::window_update(stream_id, self.config.receive_window);
+
+                if let Some(credit) = shared.next_window_update() {
+                    shared.window += credit;
+                    let mut frame = Frame::window_update(stream_id, credit);
                     frame.header_mut().ack();
                     window_update = Some(frame)
-                } else {
-                    window_update = None
                 }
             }
             if window_update.is_none() {
@@ -706,12 +702,9 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             if let Some(w) = shared.reader.take() {
                 w.wake()
             }
-            if !is_finish
-                && shared.window == 0
-                && self.config.window_update_mode == WindowUpdateMode::OnReceive
-            {
-                shared.window = self.config.receive_window;
-                let frame = Frame::window_update(stream_id, self.config.receive_window);
+            if let Some(credit) = shared.next_window_update() {
+                shared.window += credit;
+                let frame = Frame::window_update(stream_id, credit);
                 return Action::Update(frame)
             }
         } else if !is_finish {
@@ -781,10 +774,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 w.wake()
             }
         } else if !is_finish {
-            log::debug!("{}/{}: window update for unknown stream", self.id, stream_id);
-            let mut header = Header::data(stream_id, 0);
-            header.rst();
-            return Action::Reset(Frame::new(header))
+            log::debug!("{}/{}: window update for unknown stream, ignoring", self.id, stream_id);
         }
 
         Action::None
@@ -939,4 +929,3 @@ where
         }
     })
 }
-
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -713,11 +713,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             }
         } else {
             log::debug!("{}/{}: data for unknown stream, ignoring", self.id, stream_id);
-            // Previous implementations would send back a stream reset when receiving a data frame
-            // on an unknown stream. There are multiple benign scenarios that can lead to this
-            // happening and where a stream reset would not be appropriate. One such case would be
-            // when still in the process of sending out a frame closing the stream, while the stream
-            // data structure itself has already been garbage collected.
+            // We do not consider this a protocol violation and thus do not send a stream reset
+            // because we may still be processing pending `StreamCommand`s of this stream that were
+            // sent before it has been dropped and "garbage collected". Such a stream reset would
+            // interfere with the frames that still need to be sent, causing premature stream
+            // termination for the remote.
             //
             // See https://github.com/paritytech/yamux/issues/110 for details.
         }
@@ -783,11 +783,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             }
         } else {
             log::debug!("{}/{}: window update for unknown stream, ignoring", self.id, stream_id);
-            // Previous implementations would send back a stream reset when receiving a window
-            // update frame on an unknown stream. There are multiple benign scenarios that can lead
-            // to this happening and where a stream reset would not be appropriate. One such case
-            // would be when still in the process of sending out a frame closing the stream, while
-            // the stream data structure itself has already been garbage collected.
+            // We do not consider this a protocol violation and thus do not send a stream reset
+            // because we may still be processing pending `StreamCommand`s of this stream that were
+            // sent before it has been dropped and "garbage collected". Such a stream reset would
+            // interfere with the frames that still need to be sent, causing premature stream
+            // termination for the remote.
             //
             // See https://github.com/paritytech/yamux/issues/110 for details.
         }
@@ -806,11 +806,10 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             return Action::Ping(Frame::new(hdr))
         }
         log::debug!("{}/{}: ping for unknown stream", self.id, stream_id);
-        // Previous implementations would send back a stream reset when receiving a ping frame on
-        // an unknown stream. There are multiple benign scenarios that can lead to this happening
-        // and where a stream reset would not be appropriate. One such case would be when still in
-        // the process of sending out a frame closing the stream, while the stream data structure
-        // itself has already been garbage collected.
+        // We do not consider this a protocol violation and thus do not send a stream reset because
+        // we may still be processing pending `StreamCommand`s of this stream that were sent before
+        // it has been dropped and "garbage collected". Such a stream reset would interfere with the
+        // frames that still need to be sent, causing premature stream termination for the remote.
         //
         // See https://github.com/paritytech/yamux/issues/110 for details.
 
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -667,11 +667,13 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 shared.window = shared.window.saturating_sub(frame.body_len());
                 shared.buffer.push(frame.into_body());
 
-                if let Some(credit) = shared.next_window_update() {
-                    shared.window += credit;
-                    let mut frame = Frame::window_update(stream_id, credit);
-                    frame.header_mut().ack();
-                    window_update = Some(frame)
+                if matches!(self.config.window_update_mode, WindowUpdateMode::OnReceive) {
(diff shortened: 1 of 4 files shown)


romanb:
Below are just two comments on the PR description, without having looked at the diff yet (I will do that a bit later).

> Ignore WindowUpdate messages for unknown streams. With this commit
WindowUpdate messages are sent more agressively. The following scenario
would surface: A sender would close its channel, eventually being
garbage collected. The receiver, before receiving the closing message,
sends out a WindowUpdate message. The sender receives the WindowUpdate
message not being able to associate it to a stream, thus resetting the
whole connection.

Isn't the current implementation just resetting the (sub)stream [here](https://github.com/paritytech/yamux/blob/develop/src/connection.rs#L718-L721)? I think ignoring such frames is fine, but I don't currently see how it would lead to resetting the whole connection,a s you say.

> When sending very large messages, WindowUpdate messages are still not
returned to the sender in time, thus the sender still being blocked in
such case. This can be prevented by splitting large messages into
smaller Yamux frames, see
#100 (comment).
This additional optimization can be done in a future commit without
interfering with the optimization introduced in this commit.

I agree it would be a good thing to do, especially in the current state where `Connection::next_stream()` is `.await`ing the sending of each frame before doing anything else, including reading another frame from the underlying socket, so smaller frames would allow more interleaving. I'm still not sure how to go about https://github.com/paritytech/yamux/pull/104 as in general the ability to read does fundamentally depend on the ability to write (window updates at the least, but also potentially reset or close frames as a result of reading). That is, if writing stalled then reading always necessarily eventually stalls and we're only talking about whether it makes sense to delay that situation further by adding more buffers for pending frames to "extend" the socket send buffers.

mxinden:
> That is, if writing stalled then reading always necessarily eventually stalls and we're only talking about whether it makes sense to delay that situation further by adding more buffers for pending frames to "extend" the socket send buffers.

@romanb I don't follow. What do you mean with "adding more buffers for pending frames to "extend" the socket send buffers"? Just to make sure there is no confusion, in my comment above I am not suggesting to extend any buffers.

romanb:
> @romanb I don't follow. What do you mean with "adding more buffers for pending frames to "extend" the socket send buffers"? Just to make sure there is no confusion, in my comment above I am not suggesting to extend any buffers.

That was just a comment relating to my struggle with #104 in the context of saying that I agree that smaller frame sizes may be beneficial in more than just one way. Sorry for the possible confusion.

src/connection/stream.rs
```diff
@@ -175,16 +172,29 @@ impl Stream {
...
+        // When using [`WindowUpdateMode::OnReceive`] window update messages are
+        // send early on data receival (see [`crate::Connection::on_frame`]).
+        if matches!(self.config.window_update_mode, WindowUpdateMode::OnReceive) {
+            return Poll::Ready(Ok(()));
+        }
+
+        let mut shared = self.shared.lock();
+
+        if let Some(credit) = shared.next_window_update() {
             ready!(self.sender.poll_ready(cx).map_err(|_| self.write_zero_err())?);
```
romanb, line 187: If I understand correctly, then now `Stream::poll_read` and `Stream::poll_next` will effectively block reading on the sending of a window update if half of the receive window has been consumed. That seems unfortunate. How about returning `Poll::Pending` here _only if_ the `shared.window` is `0`? So if `sender.poll_ready()` gives `Poll::Pending` but the window is `> 0` we can continue, as before. Alternatively we may not need to "block" at all on `ready!(send_window_update())` within `Stream::poll_read()` and `Stream::poll_next()`.

src/connection/stream.rs
```diff
@@ -207,48 +217,33 @@ impl futures::stream::Stream for Stream {
             return Poll::Ready(None)
         }
 
-        ready!(self.send_pending_window_update(cx))?;
+        ready!(self.send_window_update(cx))?;
```
romanb, line 220: I know it was the case before and it may be unlikely to happen that `sender.poll_ready()` is `Poll::Pending`, but I'm wondering why we should ever block reading (via the `ready!` here) from the already buffered data in `Stream::poll_read` or `Stream::poll_next` on the ability to send a window update to the _bounded_ `StreamCommand` channel beforehand. It seems to potentially just create more opportunity for situations as observed in #104, if `Connection::next_stream` is currently is currently "stuck" in a `send().await`.

src/connection/stream.rs
```diff
@@ -414,5 +396,46 @@ impl Shared {
...
+        let new_credit = match self.config.window_update_mode {
+            WindowUpdateMode::OnReceive => {
+                debug_assert!(self.config.receive_window >= self.window);
+                let bytes_received = self.config.receive_window.saturating_sub(self.window);
+                bytes_received
+            },
+            WindowUpdateMode::OnRead => {
+                debug_assert!(self.config.receive_window >= self.window);
+                let bytes_received = self.config.receive_window.saturating_sub(self.window);
+                let buffer_len: u32 = self.buffer.len()
```
romanb, line 421: As far as I know `self.buffer.len()` is always a `fold()` on all chunks. Doing that essentially on every invocation of `poll_read` or `poll_next` makes me a bit uncomfortable (i.e. more frequently than it is currently done, I think once for each received data frame). Since each chunk comes from a data frame, these may be relatively small and we may even want to have smaller data frames rather than making them as large as possibly the window size, I wonder if the `Chunks` buffer should not better explicitly track its total size as chunks are `push`ed and `pop`ped.

mxinden:
>> Ignore WindowUpdate messages for unknown streams. With this commit WindowUpdate messages are sent more agressively. The following scenario would surface: A sender would close its channel, eventually being garbage collected. The receiver, before receiving the closing message, sends out a WindowUpdate message. The sender receives the WindowUpdate message not being able to associate it to a stream, thus resetting the whole connection.

> Isn't the current implementation just resetting the (sub)stream here? I think ignoring such frames is fine, but I don't currently see how it would lead to resetting the whole connection, as you say.

You are correct. My earlier assumption is incomplete. The issue at hand is a lot more subtle. See https://github.com/paritytech/yamux/issues/110. Thank you for catching this.

mxinden (commented):
Thanks @romanb for the thorough review!

src/connection/stream.rs
mxinden: > Doing that essentially on every invocation of poll_read or poll_next makes me a bit uncomfortable

I profiled a 10 second benchmark run, using 10 streams with 10_000 messages of 256 byte each on an unconstrained network. I would imagine this scenario to cause many calls to `Chunks::len`. The amount of `Chunk`s in `Chunks` alternates between 0 and ~400. Below is the [flamegraph](https://github.com/flamegraph-rs/flamegraph) svg (as a .zip as Github doesn't allow svg). Indeed the call to `fold` shows up, taking around 1% of the overall CPU time.


[flamegraph.zip](https://github.com/paritytech/yamux/files/5958408/flamegraph.zip)

Running with the same settings as above, but instead using the gbit-lan scenario the `fold` calls take less than 0.2% of the overall CPU time.

> I wonder if the Chunks buffer should not better explicitly track its total size as chunks are pushed and popped.

One could explicitly track the size of all chunks through `push` and `pop`. Problem is that `Chunks` allows mutable access to the front chunk through `front_mut` for example used [in `Stream as AsyncRead`](https://github.com/paritytech/yamux/blob/90720d796ef66b64f6453d8342c8f786ea5611ef/src/connection/stream.rs#L273-L285). One could keep the length for all chunks and subtract the offset (bytes already read) of the first chunk on each call to `Chunks::len`.

Running the same benchmark with the diff below causes the `fold` call in `Chunks::len` to no longer show up at all. I am not sure this optimization is worth the additional complexity. In case we would like to optimize this code path, I would expect a switch to a single ring-buffer per `Stream` to be more fruitful.

What do you think @romanb?

```diff
diff --git a/src/chunks.rs b/src/chunks.rs
index c22308e..2f97424 100644
--- a/src/chunks.rs
+++ b/src/chunks.rs
@@ -17,24 +17,24 @@ use std::{collections::VecDeque, io};
 /// [`Chunk`] elements.
 #[derive(Debug)]
 pub(crate) struct Chunks {
-    seq: VecDeque<Chunk>
+    seq: VecDeque<Chunk>,
+    len: usize,
 }
 
 impl Chunks {
     /// A new empty chunk list.
     pub(crate) fn new() -> Self {
-        Chunks { seq: VecDeque::new() }
+        Chunks { seq: VecDeque::new(), len: 0 }
     }
 
     /// The total length of bytes contained in all `Chunk`s.
     pub(crate) fn len(&self) -> Option<usize> {
-        self.seq.iter().fold(Some(0), |total, x| {
-            total.and_then(|n| n.checked_add(x.len()))
-        })
+        Some(self.len - self.seq.front().map(|c| c.offset()).unwrap_or(0))
     }
 
     /// Add another chunk of bytes to the end.
     pub(crate) fn push(&mut self, x: Vec<u8>) {
+        self.len += x.len();
         if !x.is_empty() {
             self.seq.push_back(Chunk { cursor: io::Cursor::new(x) })
         }
@@ -42,7 +42,9 @@ impl Chunks {
 
     /// Remove and return the first chunk.
     pub(crate) fn pop(&mut self) -> Option<Chunk> {
-        self.seq.pop_front()
+        let chunk = self.seq.pop_front();
+        self.len -= chunk.as_ref().map(|c| c.len() + c.offset()).unwrap_or(0);
+        chunk
     }
```

src/connection.rs
```diff
@@ -658,24 +658,22 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
...
-                {
-                    shared.window = self.config.receive_window;
-                    let mut frame = Frame::window_update(stream_id, self.config.receive_window);
-                    frame.header_mut().ack();
-                    window_update = Some(frame)
-                } else {
-                    window_update = None
+
+                if matches!(self.config.window_update_mode, WindowUpdateMode::OnReceive) {
+                    if let Some(credit) = shared.next_window_update() {
```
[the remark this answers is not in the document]
romanb: > I understand that the former is the context and the latter is the currently configured WindowUpdateMode

The argument given is just what you otherwise check against at the call-site, `WindowUpdateMode::OnReceive` in this case, i.e. you just leave the check against the config to `next_window_update()`. It just strikes me as a bit odd that you have to check the configured window update mode before calling `next_window_update()` when that function itself is already checking the configured mode in order to compute the update _and_ may even return `None` if there is no next window update. I thus thought it may be better to have `next_window_update()` decide a) whether there is a next window update for the given mode (returning `None` if not) and b) how to generate it, simultaneously. I guess if `next_window_update()` did not return an `Option`, making it clear that the call-site needs to decide whether to generate one, I would not have made the suggestion.  I don't really mind though, it was just a thought, I don't feel strongly about it either.

src/connection.rs
```diff
@@ -706,19 +704,22 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
...
+                    return Action::Update(frame)
+                }
+            }
+        } else {
+            log::debug!("{}/{}: data for unknown stream, ignoring", self.id, stream_id);
+            // Previous implementations would send back a stream reset when receiving a data frame
+            // on an unknown stream. There are multiple benign scenarios that can lead to this
+            // happening and where a stream reset would not be appropriate. One such case would be
+            // when still in the process of sending out a frame closing the stream, while the stream
+            // data structure itself has already been garbage collected.
```
romanb, line 720:
```suggestion
            // We do not consider this a protocol violation and thus do not send a stream reset
            /// because we may still be processing pending `StreamCommand`s of this
            /// stream that were sent before it has been dropped and "garbage collected".
            /// Such a stream reset would interfere with the frames that still need to be
            /// sent, causing premature stream termination for the remote. 
```
Just a suggestion, I don't think it is necessary to refer to previous implementations (there is the git history). Feel free to ignore. If you take the suggestion, then it should be the same in all places, of course.

src/connection/stream.rs
romanb: I did not consider the `front_mut` use, though the patch you've shown for explicit tracking of the length looks simple enough to me, i.e. I'd probably do it, but if you consider the impact shown by your measurements to be negligible, feel free to leave it for now. Thanks for the follow-up.

romanb (approved):
Looks good to me, I only left a few last suggestions. Thanks for the detailed analysis in #100. Hopefully this turns out as the expected improvement "in the wild"!

src/connection.rs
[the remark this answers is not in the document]
mxinden: Thanks for the detailed elaboration. I am torn between the two, finding both not fully intuitive. I will leave it as is, but happy to discuss alternative approaches.

src/connection/stream.rs
mxinden: > though the patch you've shown for explicit tracking of the length looks simple enough to me

Sounds good then. Done in https://github.com/paritytech/yamux/pull/109/commits/d4398f0fa23870e66a3e2a69fb99092a3cfb1e77.

src/connection.rs
mxinden: Better for sure. Done with https://github.com/paritytech/yamux/pull/109/commits/55db98a5133b5c563ab56bb7b06338d57bdca168.

romanb:
Feel free to merge, or let me know if I should so so.

PR #112, opened 2021-02-11
Allow reads concurrent to pending writes.

romanb:
Currently, all write operations (e.g. for a particular frame sent by a substream) are `.await`ed for completion within the `Connection::next()` loop, thus prohibiting reading from the underlying socket while the write is pending. This can lead to unexpected write deadlocks between two peers in specific, though likely uncommon, situations. See  https://github.com/paritytech/yamux/pull/104 for further details and a test.

This PR proposes to allow concurrent reads while a frame write is pending, thereby propagating pending write back-pressure via the paused command channels, rather than doing so by "blocking"`Connection::next()` entirely until such a write completes.

Notably, no new buffers are introduced. We still send one frame at a time from the context of a `Connection`. However, when a frame write cannot complete, the command channels are paused and I/O reads can continue while the write is pending. The paused command channels exercise write back-pressure on the streams and API and ensure that the only frames we still try to send are those as a result of reading a frame - these writes then indeed wait for completion of the prior pending send operation, if any, but since it is done as a result of reading a frame, the remote can in turn usually write again, should it have been waiting to be able to do so before it in turn can read again. Unexpected write deadlocks of peers which otherwise appear to read & write concurrently from substreams can thus be avoided. I'm not entirely sure if there may not still be situations where a write deadlock can be provoked but I did not want to introduce new "pending frames" buffers, which would again be bounded and raise the question of what to do when these are also full. I thus came up with this "refinement" of the status quo without new buffers.

The previously ignored test introduced in #104 is hereby enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ static_assertions = "1"
 anyhow = "1"
 criterion = "0.3"
 futures = "0.3.4"
-quickcheck = "0.9"
+quickcheck = "1.0"
 tokio = { version = "0.2", features = ["tcp", "rt-threaded", "macros"] }
 tokio-util = { version = "0.3", features = ["compat"] }
 constrained-connection = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-futures = { version = "0.3.4", default-features = false, features = ["std"] }
+futures = { version = "0.3.12", default-features = false, features = ["std"] }
 log = "0.4.8"
 nohash-hasher = "0.2"
 parking_lot = "0.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ futures = { version = "0.3.4", default-features = false, features = ["std"] }
 log = "0.4.8"
 nohash-hasher = "0.2"
 parking_lot = "0.11"
-rand = "0.7.2"
+rand = "0.8.3"
 static_assertions = "1"
 
 [dev-dependencies]
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -395,14 +395,14 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 } else {
                     let socket = &mut self.socket;
                     let io = future::poll_fn(move |cx| {
-                        // Progress writing.
-                        match socket.get_mut().poll_ready_unpin(cx) {
-                            Poll::Pending => {}
-                            Poll::Ready(res) => {
+                        // If the stream command receiver was paused, we got
+                        // back-pressure from the underlying I/O stream and
+                        // must wait for the sink to be ready again for the
+                        // next frame.
+                        if stream_receiver_paused {
+                            if let Poll::Ready(res) = socket.get_mut().poll_ready_unpin(cx) {
                                 res.or(Err(ConnectionError::Closed))?;
-                                if stream_receiver_paused {
-                                    return Poll::Ready(Result::Ok(IoEvent::OutboundReady))
-                                }
+                                return Poll::Ready(Result::Ok(IoEvent::OutboundReady))
                             }
                         }
                         // Progress reading.
@@ -460,6 +460,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
 
             match io_event? {
                 Poll::Ready(IoEvent::OutboundReady) => {
+                    log::debug!("{}: Sink ready, unpausing command streams.", self.id);
                     self.stream_receiver.unpause();
                     // Only unpause the control command receiver if we're not
                     // shutting down already.
@@ -756,7 +757,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 }
             }
         } else {
-            log::debug!("{}/{}: data for unknown stream, ignoring", self.id, stream_id);
+            log::trace!("{}/{}: data frame for unknown stream, possibly dropped earlier: {:?}",
+                self.id, stream_id, frame);
             // We do not consider this a protocol violation and thus do not send a stream reset
             // because we may still be processing pending `StreamCommand`s of this stream that were
             // sent before it has been dropped and "garbage collected". Such a stream reset would
@@ -826,7 +828,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 w.wake()
             }
         } else {
-            log::debug!("{}/{}: window update for unknown stream, ignoring", self.id, stream_id);
+            log::trace!("{}/{}: window update for unknown stream, possibly dropped earlier: {:?}",
+                self.id, stream_id, frame);
             // We do not consider this a protocol violation and thus do not send a stream reset
             // because we may still be processing pending `StreamCommand`s of this stream that were
             // sent before it has been dropped and "garbage collected". Such a stream reset would
@@ -849,7 +852,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
             hdr.ack();
             return Action::Ping(Frame::new(hdr))
         }
-        log::debug!("{}/{}: ping for unknown stream", self.id, stream_id);
+        log::trace!("{}/{}: ping for unknown stream, possibly dropped earlier: {:?}",
+            self.id, stream_id, frame);
         // We do not consider this a protocol violation and thus do not send a stream reset because
         // we may still be processing pending `StreamCommand`s of this stream that were sent before
         // it has been dropped and "garbage collected". Such a stream reset would interfere with the
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -400,7 +400,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                         // must wait for the sink to be ready again for the
                         // next frame.
                         if stream_receiver_paused {
-                            if let Poll::Ready(res) = socket.get_mut().poll_ready_unpin(cx) {
+                            if let Poll::Ready(res) = socket.poll_ready_unpin(cx) {
                                 res.or(Err(ConnectionError::Closed))?;
                                 return Poll::Ready(Result::Ok(IoEvent::OutboundReady))
                             }
@@ -460,7 +460,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
 
             match io_event? {
                 Poll::Ready(IoEvent::OutboundReady) => {
-                    log::debug!("{}: Sink ready, unpausing command streams.", self.id);
+                    log::trace!("{}: Sink ready, unpausing command streams.", self.id);
                     self.stream_receiver.unpause();
                     // Only unpause the control command receiver if we're not
                     // shutting down already.
@@ -628,7 +628,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                         return Ok(Some(stream))
                     }
                     Action::Update(f) => {
-                        log::trace!("{}/{}: sending update", self.id, f.header().stream_id());
+                        log::trace!("{}: sending update: {:?}", self.id, f.header());
                         self.send(f).await.or(Err(ConnectionError::Closed))?
                     }
                     Action::Ping(f) => {
@@ -958,7 +958,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 frame
             };
             if let Some(f) = frame {
-                log::trace!("{}/{}: sending final frame: {}", self.id, stream_id, f.header());
+                log::trace!("{}/{}: sending: {}", self.id, stream_id, f.header());
                 send(conn_id, &mut self.socket, &mut self.stream_receiver,
                     &mut self.control_receiver, f).await.or(Err(ConnectionError::Closed))?
             }
@@ -1020,7 +1020,7 @@ async fn send<T: AsyncRead + AsyncWrite + Unpin>(
     future::poll_fn(|cx| {
         match io.get_mut().poll_ready_unpin(cx)? {
             Poll::Pending => {
-                log::debug!("{}: send: Write pending. Continuing with paused command streams.", id);
+                log::trace!("{}: send: Write pending. Continuing with paused command streams.", id);
                 stream_receiver.pause();
                 control_receiver.pause();
                 return Poll::Ready(Result::Ok(()))
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -103,7 +103,8 @@ use futures::{
     channel::{mpsc, oneshot},
     future::{self, Either},
     prelude::*,
-    stream::{Fuse, FusedStream}
+    stream::{Fuse, FusedStream},
+    sink::SinkExt,
 };
 use nohash_hasher::IntMap;
 use std::{fmt, io, sync::Arc, task::{Context, Poll}};
@@ -395,9 +396,9 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     let socket = &mut self.socket;
                     let io = future::poll_fn(move |cx| {
                         // Progress writing.
-                        match socket.get_mut().poll_send::<()>(cx, None) {
-                            frame::PollSend::Pending(_) => {}
-                            frame::PollSend::Ready(res) => {
+                        match socket.get_mut().poll_ready_unpin(cx) {
+                            Poll::Pending => {}
+                            Poll::Ready(res) => {
                                 res.or(Err(ConnectionError::Closed))?;
                                 if stream_receiver_paused {
                                     return Poll::Ready(Result::Ok(IoEvent::OutboundReady))
@@ -883,7 +884,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
     /// Try to flush the underlying I/O stream, without waiting for it.
     async fn flush_nowait(&mut self) -> Result<()> {
         future::poll_fn(|cx| {
-            let _ = self.socket.get_mut().poll_flush(cx)?;
+            let _ = self.socket.get_mut().poll_flush_unpin(cx)?;
             Poll::Ready(Ok(()))
         }).await
     }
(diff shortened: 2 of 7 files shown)


romanb:
> I'm not entirely sure if there may not still be situations where a write deadlock can be provoked but I did not want to introduce new "pending frames" buffers,  [..]

To be fair, one can say that I essentially introduced a buffer of 1 on top of the send buffer / capacity of the connection in order to break this particular deadlock. The question then is whether that is worthwhile doing and can be considered a small enhancement, or is just more complexity for too little gain. Fundamentally, one eventually needs to either wait/block on sending or abort the connection if writes cannot proceed, as it is not possible to only read (i.e. at least window updates must be sent).

mxinden (commented):
I like the general structure of the approach taken here, namely to stop polling the stream- and command-receiver when the socket is busy sending an item, as well as introducing a one-item send buffer  into `Io`.

It did take me a bit of time to fully understand the concepts introduced in this pull request (`PollSend`) as well as how they interact with the existing `Pausable` wrapper.

I have not yet put enough thoughts into the suggestions below to consider them a solid review, thus please take the below with a grain of salt. I have played with your changes locally, but I am not yet at a state where my local changes compile. 

- How about implementing `Sink` for `Io`, keeping your new "buffer of 1 on top of the send buffer".
- How about in `Connection::next` only poll `self.stream_receiver` and `self.command_receiver` when `self.socket.poll_ready` returns `Poll::Ready(Ok())`?
- How about in `Connection::on_control_command`, `Connection::on_stream_command` and `Connection::on_frame` to [use `SinkExt::feed`](https://docs.rs/futures/0.3.12/futures/prelude/sink/trait.SinkExt.html#method.feed) instead of the custom `send` function?

I would expect using the concepts from the `futures` library, namely `Sink` and `SinkExt`, instead of our own `PollSend` and `Pausable` to make this pull request easier to understand, given that one is likely already familiar with the former concepts.

What do you think? Happy to expand on my ideas, or try to provide a compiling version.



<details>
  <summary>My work-in-progress changes on top of this pull request</summary>


```diff
diff --git a/src/connection.rs b/src/connection.rs
index 60bed2c..355af6c 100644
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -392,27 +392,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     num_terminated += 1;
                     Either::Left(future::pending())
                 } else {
-                    let socket = &mut self.socket;
-                    let io = future::poll_fn(move |cx| {
-                        // Progress writing.
-                        match socket.get_mut().poll_send::<()>(cx, None) {
-                            frame::PollSend::Pending(_) => {}
-                            frame::PollSend::Ready(res) => {
-                                res.or(Err(ConnectionError::Closed))?;
-                                if stream_receiver_paused {
-                                    return Poll::Ready(Result::Ok(IoEvent::OutboundReady))
-                                }
-                            }
-                        }
-                        // Progress reading.
-                        let next_frame = match futures::ready!(socket.poll_next_unpin(cx)) {
-                            None => Ok(None),
-                            Some(Err(e)) => Err(e.into()),
-                            Some(Ok(f)) => Ok(Some(f))
-                        };
-                        Poll::Ready(Ok(IoEvent::Inbound(next_frame)))
-                    });
-                    Either::Right(io)
+                    Either::Right(self.socket.try_next().err_into())
                 };
 
             let mut next_stream_command =
@@ -420,7 +400,19 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     num_terminated += 1;
                     Either::Left(future::pending())
                 } else {
-                    Either::Right(self.stream_receiver.next())
+                    Either::Right(
+                        async {
+                            // mxinden: Only consider polling the stream_receiver when the socket is ready to
+                            // accept a new item.
+                            //
+                            // mxinden: Likely does not pass the borrow checker with the socket
+                            // being borrowed both here and below for the next_control_command.
+                            future::poll_fn(|cx| {
+                                self.socket.poll_ready(cx)
+                            }).await;
+                            self.stream_receiver.next()
+                        }
+                    )
                 };
 
             let mut next_control_command =
@@ -428,7 +420,18 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     num_terminated += 1;
                     Either::Left(future::pending())
                 } else {
-                    Either::Right(self.control_receiver.next())
+                    Either::Right(
+                        async {
+                            // mxinden: Only consider polling the control_receiver when the socket is ready
+                            // to accept a new item.
+                            future::poll_fn(|cx| {
+                            future::poll_fn(|cx| {
+                                self.socket.poll_ready(cx)
+                            }).await;
+                            self.control_receiver.next()
+
+                        }
+                    )
                 };
 
             if num_terminated == 3 {
@@ -504,7 +507,9 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                     let mut frame = Frame::window_update(id, extra_credit);
                     frame.header_mut().syn();
                     log::trace!("{}/{}: sending initial {}", self.id, id, frame.header());
-                    self.send(frame).await.or(Err(ConnectionError::Closed))?
+                    // mxinden: Use SinkExt::feed which will send the item into the sink, but does
+                    // not consecutively flush it.
+                    self.socket.get_mut().feed(frame).await.or(Err(ConnectionError::Closed))?
                 }
                 let stream = {
                     let config = self.config.clone();
@@ -557,10 +562,6 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
         Ok(())
     }
 
-    async fn send(&mut self, f: impl Into<Frame<()>>) -> Result<()> {
-        send(self.id, &mut self.socket, &mut self.stream_receiver, &mut self.control_receiver, f).await
-    }
-
     /// Process a command from one of our `Stream`s.
     async fn on_stream_command(&mut self, cmd: Option<StreamCommand>) -> Result<()> {
         match cmd {
@@ -996,54 +997,6 @@ enum IoEvent {
     OutboundReady,
 }
 
-/// Sends a frame on the given `io` stream.
-///
-/// If the frame is taken by `io` but cannot be fully sent, the command
-/// receivers are paused, without waiting for completion.
-///
-/// If a prior send operation is still pending, waits for its completion.
-async fn send<T: AsyncRead + AsyncWrite + Unpin>(
-    id: Id,
-    io: &mut Fuse<frame::Io<T>>,
-    stream_receiver: &mut Pausable<mpsc::Receiver<StreamCommand>>,
-    control_receiver: &mut Pausable<mpsc::Receiver<ControlCommand>>,
-    frame: impl Into<Frame<()>>
-) -> Result<()> {
-    let mut frame = Some(frame.into());
-    future::poll_fn(move |cx| {
-        let next = frame.take().expect("Frame has not yet been taken by `io`.");
-        match io.get_mut().poll_send(cx, Some(next)) {
-            frame::PollSend::Pending(Some(f)) => {
-                debug_assert!(stream_receiver.is_paused());
-                log::debug!("{}: send: Prior write pending. Waiting.", id);
-                frame = Some(f);
-                return Poll::Pending
-            }
-            frame::PollSend::Pending(None) => {
-                // The frame could not yet fully be written to the underlying
-                // socket, so we pause the processing of commands in order to
-                // pause writing while still being able to read from the socket.
-                // The only frames that may still be sent while commands are paused
-                // are as a reaction to frames being read, which in turn allows
-                // the remote to make progress eventually, if it should
-                // currently be blocked on writing. In this way unnecessary
-                // deadlocks between peers blocked on writing are avoided.
-                log::trace!("{}: send: Write pending. Continuing with paused command streams.", id);
-                stream_receiver.pause();
-                control_receiver.pause();
-                return Poll::Ready(Ok(()))
-            }
-            frame::PollSend::Ready(Err(e)) => {
-                return Poll::Ready(Err(e.into()))
-            }
-            frame::PollSend::Ready(Ok(())) => {
-                // Note: We leave the unpausing of the command streams to `Connection::next()`.
-                return Poll::Ready(Ok(()))
-            }
-        }
-    }).await
-}
-
 /// Turn a Yamux [`Connection`] into a [`futures::Stream`].
 pub fn into_stream<T>(c: Connection<T>) -> impl futures::stream::Stream<Item = Result<Stream>>
 where
diff --git a/src/frame/io.rs b/src/frame/io.rs
index f909aec..13f6f4f 100644
--- a/src/frame/io.rs
+++ b/src/frame/io.rs
@@ -34,45 +34,48 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Io<T> {
         }
     }
 
-    /// Continue writing on the underlying connection, possibly with a new frame.
-    ///
-    /// If there is no data pending to be written from a prior frame, an attempt
-    /// is made at writing the new frame, if any.
-    ///
-    /// If there is data pending to be written from a prior frame, an attempt is made to
-    /// complete the pending writes before accepting the new frame, if any. If
-    /// the new frame cannot be accepted due to pending writes, it is returned.
-    pub(crate) fn poll_send<A>(&mut self, cx: &mut Context<'_>, mut frame: Option<Frame<A>>)
-        -> PollSend<A, io::Result<()>>
-    {
+    #[cfg(test)]
+    async fn flush(&mut self) -> io::Result<()> {
+        future::poll_fn(|cx| self.poll_flush(cx)).await
+    }
+
+    #[cfg(test)]
+    async fn send<A>(&mut self, frame: Frame<A>) -> io::Result<()> {
+        let mut frame = Some(frame);
+        future::poll_fn(|cx| {
+            match self.poll_send(cx, Some(frame.take().expect("`frame` not yet taken"))) {
+                PollSend::Pending(f) => { frame = f; return Poll::Pending }
+                PollSend::Ready(r) => return Poll::Ready(r)
+            }
+        }).await
+    }
+
+    pub(crate) async fn close(&mut self) -> io::Result<()> {
+        future::poll_fn(|cx| self.poll_flush(cx)).await?;
+        self.io.close().await
+    }
+}
+
+impl <T: AsyncWrite + Unpin, A> Sink<Frame<A>> for Io<T> {
+    type Error = io::Error;
+
+    fn poll_ready(
+        self: Pin<&mut Self>,
+        cx: &mut Context
+    ) -> Poll<Result<(), Self::Error>> {
         loop {
             match &mut self.write_state {
                 WriteState::Init => {
-                    if let Some(f) = frame.take() {
-                        let header = header::encode(&f.header);
-                        let buffer = f.body;
-                        self.write_state = WriteState::Header { header, buffer, offset: 0 };
-                    } else {
-                        return PollSend::Ready(Ok(()))
-                    }
+                    return Poll::Ready(Ok(()))
                 }
                 WriteState::Header { header, buffer, ref mut offset } => {
-                    match Pin::new(&mut self.io).poll_write(cx, &header[*offset ..]) {
-                        Poll::Pending => {
-                            if let Some(f) = frame.take() {
-                                return PollSend::Pending(Some(f))
-                            }
-                            return PollSend::Pending(None)
-                        }
-                        Poll::Ready(Err(e)) => {
-                            if let Some(f) = frame.take() {
-                                return PollSend::Pending(Some(f))
-                            }
+                    match ready!(Pin::new(&mut self.io).poll_write(cx, &header[*offset ..])) {
+                        Err(e) => {
                             return PollSend::Ready(Err(e))
                         }
-                        Poll::Ready(Ok(n)) => {
+                        Ok(n) => {
                             if n == 0 {
-                                return PollSend::Ready(Err(io::ErrorKind::WriteZero.into()))
+                                return Poll::Ready(Err(io::ErrorKind::WriteZero.into()))
                             }
                             *offset += n;
                             if *offset == header.len() {
@@ -87,20 +90,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Io<T> {
                     }
                 }
                 WriteState::Body { buffer, ref mut offset } => {
-                    match Pin::new(&mut self.io).poll_write(cx, &buffer[*offset ..]) {
-                        Poll::Pending => {
-                            if let Some(f) = frame.take() {
-                                return PollSend::Pending(Some(f))
-                            }
-                            return PollSend::Pending(None)
+                    match ready!(Pin::new(&mut self.io).poll_write(cx, &buffer[*offset ..])) {
+                        Err(e) => {
+                            return Poll::Ready(Err(e))
                         }
-                        Poll::Ready(Err(e)) => {
-                            if let Some(f) = frame.take() {
-                                return PollSend::Pending(Some(f))
-                            }
-                            return PollSend::Ready(Err(e))
-                        }
-                        Poll::Ready(Ok(n)) => {
+                        Ok(n) => {
                             if n == 0 {
                                 return PollSend::Ready(Err(io::ErrorKind::WriteZero.into()))
                             }
@@ -115,33 +109,42 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Io<T> {
         }
     }
 
-    pub(crate) fn poll_flush(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
-        match self.poll_send(cx, Option::<Frame<()>>::None) {
-            PollSend::Pending(_) => return Poll::Pending,
-            PollSend::Ready(r) => r?
+    /// Continue writing on the underlying connection, possibly with a new frame.
+    ///
+    /// If there is no data pending to be written from a prior frame, an attempt
+    /// is made at writing the new frame, if any.
+    ///
+    /// If there is data pending to be written from a prior frame, an attempt is made to
+    /// complete the pending writes before accepting the new frame, if any. If
+    /// the new frame cannot be accepted due to pending writes, it is returned.
+    pub(crate) fn start_send(self: Pin<&mut Self>, mut f: Frame<A>)
+        -> Result<(), Self::Error>
+    {
+        loop {
+            match &mut self.write_state {
+                WriteState::Init => {
+                    let header = header::encode(&f.header);
+                    let buffer = f.body;
+                    self.write_state = WriteState::Header { header, buffer, offset: 0 };
+                }
+                WriteState::Header { header, buffer, ref mut offset } => {
+                    panic!();
+                }
+                WriteState::Body { buffer, ref mut offset } => {
+                    panic!();
+                }
+            }
         }
-        Pin::new(&mut self.io).poll_flush(cx)
-    }
-
-    pub(crate) async fn close(&mut self) -> io::Result<()> {
-        future::poll_fn(|cx| self.poll_flush(cx)).await?;
-        self.io.close().await
     }
 
-    #[cfg(test)]
-    async fn flush(&mut self) -> io::Result<()> {
-        future::poll_fn(|cx| self.poll_flush(cx)).await
+    pub(crate) fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        self.poll_ready(cx)
     }
-
-    #[cfg(test)]
-    async fn send<A>(&mut self, frame: Frame<A>) -> io::Result<()> {
-        let mut frame = Some(frame);
-        future::poll_fn(|cx| {
-            match self.poll_send(cx, Some(frame.take().expect("`frame` not yet taken"))) {
-                PollSend::Pending(f) => { frame = f; return Poll::Pending }
-                PollSend::Ready(r) => return Poll::Ready(r)
-            }
-        }).await
+    fn poll_close(
+        self: Pin<&mut Self>,
+        cx: &mut Context
+    ) -> Poll<Result<(), Self::Error>> {
+        unimplemented!();
     }
 }
```
 
</details>

romanb:
Thanks for the review. I will take a closer look at your suggestions and report back here. I also want to give this some morethought and testing. Meanwhile, I think it makes sense to release the current state in `develop`, so I will prepare a PR for such a release today.

romanb:
> My work-in-progress changes on top of this pull request.

The main problem with your changes in `connection.rs` is that you cannot mutably borrow `self.socket` across multiple futures like this, which is why at least this particular approach seems to be a dead end.

romanb:
As a first improvement based on your suggestions, `Io` now implements `Sink` in https://github.com/paritytech/yamux/pull/112/commits/eb6c73201822d2835320414b133ad6c36a5bbafb.

romanb:
A small update: The write deadlock seems to be relatively easy to reproduce with TCP in the context of the test in `tests/concurrent.rs` by playing around with the TCP send buffer sizes, which I did on `master`, `develop` as well as this branch for comparison. To be clear upfront though, and as already stated earlier, the changes in this PR cannot completely avoid a mutual write deadlock under any circumstance - at the end of the day, one needs to tune the TCP send buffer sizes according to the expected throughput and traffic pattern. Nevertheless, the change in this PR to continue reading from the socket when the sending of the current frame is `Pending` seems to reduce the likelihood of running into such a mutual blocking write significantly. This is the case for both `WindowUpdateMode`s but special attention should be given to the fact that with `WindowUpdateMode::OnRead`, all window updates are sent from the context of a `Stream` via `StreamCommand`s. That means once the stream command channel is paused due to the current write peing `Pending` in the `Connection`, we also have back-pressure for window updates and the only remaining cases where the `Connection` may try to send a frame while continuing to read is when encountering protocol violations. So with `WindowUpdateMode::OnRead`, a mutual write deadlock is very unlikely to occur (unless the client code is at fault by not reading from streams until a large write has completed), which my testing with `tests/concurrent.rs` confirmed even with send buffer sizes that are smaller than a single Yamux frame, i.e. even that works on this branch without issues. With `WindowUpdateMode::OnReceive` a mutual write deadlock can still occur primarily due to the intermittent sending of window updates from `Connection::next_stream()`, which must await the sending of the previous frame, if any.

For anyone reading this without following all the details, I'd like to emphasise though again that even with the latest release a mutual write deadlock is unlikely to be encountered in practice as it requires sending large amounts of data simultaneously from both endpoints, coupled with insufficiently sized TCP send buffers, and increasing TCP send buffers can always solve the problem. Still, I am convinced as a result of further testing that the changes proposed here, namely to pause command streams and continue reading from the socket while the current frame write is pending, make it even more unlikely to occur and hence are beneficial and desirable.

mxinden (commented):
Thanks for the additional refactorings.

I would like to suggest one additional change, namely the removal of the `Pausable` wrapper on the stream command channel. Related to that suggestion, already raised earlier, we ran into the issue of multiple borrows of `socket` (see [comment](https://github.com/paritytech/yamux/pull/112#issuecomment-779113659)).

I found a way around the borrow issue. https://github.com/romanb/yamux/pull/1 implements the suggestion. It compiles and passes all tests, including the deadlock test. I hope for the patch to be self-explanatory. Please let me know in case it isn't.

What do you think @romanb? Would this be a simplification?

mxinden (approved):
Thanks, especially for the thorough testing.

mxinden:
As far as I can tell this pull request is ready to be merged. @tomaka would you mind approving the pull request so it can be merged?

tomaka (approved):

PR #115, opened 2021-02-16
v0.8.1

romanb:
To avoid surprises, it seems best to backport the part of `v0.9` that ignores received frames for unknown streams into a `v0.8.1` release. Otherwise nodes running `v0.8` may react with a stream reset if they get a window update from a node running `v0.9` while the stream is already closing. Rolling out `v0.8.1` first should avoid such hiccups.

cc @mxinden

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,9 +2,9 @@ name: Rust
 
 on:
   push:
-    branches: [ develop ]
+    branches: [ develop, master ]
   pull_request:
-    branches: [ develop ]
+    branches: [ develop, master ]
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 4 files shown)


tomaka (approved):

tomaka (approved):
I haven't reviewed the logic, but if nobody else reviews I can give you an approval

mxinden (approved):
> To avoid surprises, it seems best to backport the part of v0.9 that ignores received frames for unknown streams into a v0.8.1 release. Otherwise nodes running v0.8 may react with a stream reset if they get a window update from a node running v0.9 while the stream is already closing. Rolling out v0.8.1 first should avoid such hiccups.

Yes, good point! :+1:

Issue #117, opened 2021-04-15
GC Bug


macrocan:
Firstly, show one case
```cpp
    let data = b"hello world";
    stream.write_all(data).await.expect("C write msg 1");
    stream.write_all(data).await.expect("C write msg 2");
    // msg3
    stream.write_all(data).await.expect("C write msg 3");

    // drop it immediately
    drop(stream);
```
Drop stream immediately after write message. As result, msg3 can't be received by reader.
After debug, We found that GC send RST frame before msg3.
```cpp
// garbage_collect()

    if let Some(f) = frame {
        log::trace!("{}: sending: {}", self.id, f.header());
        self.socket.get_mut().send(&f).await.or(Err(ConnectionError::Closed))?
    }
```
GC send RST frame directly to reader, At the meantime, msg3 is still queueed in stream receiver.
If GC send RST frame to stream receiver, the bug will be fixed.
````cpp
// garbage_collect()

    if let Some(f) = frame {
        log::trace!("{}/{}: sending: {}", self.id, stream_id, f.header());
        // send to message process queue to process(/send) orderly
        let cmd = StreamCommand::SendFrame(f.left());
        if self.stream_sender.try_send(cmd).is_err() {
            continue;
        }
    }
````

mxinden:
Yes, indeed, there is a race condition here. Users of `yamux` have to call `poll_close` on a stream before dropping it to make sure all messages are sent before closing the stream with the remote.

Let me know in case that works for you.

macrocan:
> Yes, indeed, there is a race condition here. Users of `yamux` have to call `poll_close` on a stream before dropping it to make sure all messages are sent before closing the stream with the remote.
> 
> Let me know in case that works for you.

For convenience of user, It seems that GC can close stream automatically when stream is dropped.

tomaka:
By sending RST to the remote, you inform of the fact that the messages that the remote has sent on the substream might not have been received and processed. Sending a close message and immediately discarding the substream would be against that idea.

macrocan:
> By sending RST to the remote, you inform of the fact that the messages that the remote has sent on the substream might not have been received and processed. Sending a close message and immediately discarding the substream would be against that idea.

My mistake, when stream is dropped, Gc should reset stream after all queued message were sent.

tomaka:
[...]

PR #118, opened 2021-04-15
fix gc bug

macrocan:
#117


mxinden (commented):
I am pretty sure a fix for the _drop stream without closing it_ problem would be more involved.

For now, I would suggest making sure you call `poll_close` on the stream.

I would very much appreciate a pull request making this more clear in the docs.

In case you would like to fix this on a more fundamental level, please open up another issue with a detailed plan how to tackle this.

PR #120, opened 2021-06-03
src/lib: Default to WindowUpdateMode::OnRead

mxinden:
Default to `WindowUpdateMode::OnRead`, thus enabling full Yamux
flow-control, exercising back pressure on senders, preventing stream
resets due to reaching the buffer limit.

See the [`WindowUpdateMode` documentation] for details, especially the
section on deadlocking when sending data larger than the receivers
window.

[`WindowUpdateMode` documentation]: https://docs.rs/yamux/0.9.0/yamux/enum.WindowUpdateMode.html

This would prevent surprises like https://github.com/libp2p/rust-libp2p/issues/2089.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+# 0.10.0 [unreleased]
+
+- Default to `WindowUpdateMode::OnRead`, thus enabling full Yamux flow-control,
+  exercising back pressure on senders, preventing stream resets due to reaching
+  the buffer limit.
+
+  See the [`WindowUpdateMode` documentation] for details, especially the section
+  on deadlocking when sending data larger than the receivers window.
+
+  [`WindowUpdateMode` documentation]: https://docs.rs/yamux/0.9.0/yamux/enum.WindowUpdateMode.html
+
 # 0.9.0
 
 - Force-split larger frames, for better interleaving of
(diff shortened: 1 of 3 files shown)


AgeManning (approved):
I'm not very familiar with the intricacies here. 

It does seem like in most cases the deadlock risk is unlikely compared the benefits in setting this to default, so makes sense to me.

mxinden:
//CC @ec2 and @olibearo as you are currently using `YamuxConfig::default()`.

https://github.com/ChainSafe/forest/blob/main/node/forest_libp2p/src/service.rs#L381


//CC @koivunej for rust-ipfs:

https://github.com/rs-ipfs/rust-ipfs/blob/master/src/p2p/transport.rs#L31

//CC @dvc94ch for ipfs-embed:

https://github.com/ipfs-rust/ipfs-embed/blob/master/src/net/mod.rs#L93

koivunej (approved):
Thanks for the ping @mxinden! Per [your explanation](https://github.com/libp2p/rust-libp2p/issues/2089#issuecomment-852935108) this sounds very much a reasonable default but I don't have knowledge on yamux internals except for what I've learned by reading random issues.

dvc94ch:
Seems ok. It's not a breaking change as it only affects the receiver behavior if I understand correctly. Thanks!

PR #130, opened 2022-02-17
Fix connection flush

appaquet:
See the full conversation here: https://github.com/libp2p/rust-libp2p/issues/2461

The bump to Yamux 0.10 in libp2p broke the wasm-ext transport (used for WebSocket transport in WebAssembly). It turns out that the wasm-ext implementation makes it more prone to yield a `Pending` on a `poll_next` (see [this](https://github.com/libp2p/rust-libp2p/blob/96dbfcd1ade6de5b4ae623f59fb0d67b2916576a/transports/wasm-ext/src/lib.rs#L485)). When used in conjunction with Noise, a frame may get buffered and only written to the underlying I/O stream when flushing it (see [this](https://github.com/libp2p/rust-libp2p/blob/b79fd02f0bdde07f437c691e278a16cf9024036a/transports/noise/src/io.rs#L138)) 

In https://github.com/libp2p/rust-yamux/pull/112, flushing is no longer awaited, but instead, is called at each iteration of the `next` loop in `Connection` (see [this](https://github.com/libp2p/rust-yamux/blob/9bdebeed794e6a8498db71b50f6d543f4f1bf3c2/src/connection.rs#L498)). Unfortunately, when using wasm-ext with Noise, frames may never get written until the next iteration is triggered by another event (incoming frame or stream/control events).

As discussed in libp2p's issue, a fix is to make sure that flushing the I/O stream gets progressed in the main loop. I also think that `flush_nowait` is not required anymore since flushing is now actively done in the loop.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -431,7 +431,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 num_terminated += 1;
                 Either::Left(future::pending())
             } else {
-                // poll socket for next incoming frame, but also make sure any pending writes are properly flushed
+                // Poll socket for next incoming frame, but also make sure any pending writes are properly flushed.
                 let socket = &mut self.socket;
                 let next_frame = future::poll_fn(move |cx| {
                     if let Poll::Ready(res) = socket.poll_next_unpin(cx) {
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -408,7 +408,6 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
 
             if let IoEvent::Inbound(frame) = next_io_event.await? {
                 if let Some(stream) = self.on_frame(frame).await? {
-                    self.flush_nowait().await.or(Err(ConnectionError::Closed))?;
                     return Ok(Some(stream));
                 }
                 continue; // The socket sink still has a pending write.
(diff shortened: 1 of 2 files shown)


src/connection.rs
```diff
@@ -432,7 +431,21 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
...
             } else {
-                Either::Right(self.socket.next())
+                // poll socket for next incoming frame, but also make sure any pending writes are properly flushed
+                let socket = &mut self.socket;
+                let next_frame = future::poll_fn(move |cx| {
+                    if let Poll::Ready(res) = socket.poll_next_unpin(cx) {
+                        return Poll::Ready(res);
+                    }
+
+                    if let Poll::Ready(Err(err)) = socket.poll_flush_unpin(cx) {
```
mxinden, line 441: I guess we assume that for all implementations of `Sink` `poll_flush_unpin` is cheap? If not, we could stop calling `poll_flush_unpin` once it returns `Poll::Ready(Ok(()))`.

src/connection.rs
```diff
@@ -432,7 +431,21 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
                 num_terminated += 1;
                 Either::Left(future::pending())
             } else {
-                Either::Right(self.socket.next())
+                // poll socket for next incoming frame, but also make sure any pending writes are properly flushed
```
mxinden, line 434:
```suggestion
                // Poll socket for next incoming frame, but also make sure any pending writes are properly flushed.
```

mxinden (commented):
Again, thank you for digging so deep into this!

src/connection.rs
appaquet: I think it makes sense. But we need to make sure to reset every time we write to the underlying IO

appaquet:
@mxinden Updated. I don't know if you had something else in mind, but it does the trick.

mxinden (approved):
Thanks @appaquet. I will prepare a release candidate to test on larger deployments.

PR #137, opened 2022-08-01
src/connection: Return error from next_frame Future early

mxinden:
> The `frame` future might be _ready_ with an `Error` from the underlying
socket (i.e. here `libp2p-websocket`). Though given that the result of the
`control_command` `Future` is handled first, `on_control_command` is called
despite `frame` having returned an `Error`. `on_control_command` itself may try
to write to the underlying socket, which will panic given that the socket
returned an error earlier via the `frame` `Future`.

Patch to validate suspicion in
https://github.com/libp2p/rust-libp2p/issues/2598.

Note that I am not planning to merge this as is. This needs more refactoring in order to make it into `master`.

I would very much appreciate help testing this patch. You can add the override below to your `Cargo.toml` to compile your binary with the patch:

``` toml
[patch.crates-io]
yamux = { git = 'https://github.com/mxinden/yamux', branch = "next-frame-error" }
```

Please double check your `Cargo.lock` to make sure the patched version is being used to compile your binary.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -489,6 +489,26 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
 
             let (stream_command, control_command, frame) = next_item.await;
 
+            // Return early in case we hit an error.
+            let frame = match frame {
+                Poll::Ready(Some(Err(e))) => {
+                    let e: ConnectionError = e.into();
+                    if e.io_kind() == Some(io::ErrorKind::ConnectionReset) {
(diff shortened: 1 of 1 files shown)


jasl:
the patch should be

```
[patch.crates-io]
yamux = { git = 'https://github.com/mxinden/yamux', branch = "next-frame-error" }
```

mxinden:
Correct. Thanks @jasl. Updated above.

mxinden:
Closing here in favor of https://github.com/libp2p/rust-yamux/pull/138.

PR #138, opened 2022-08-04
src/connection: Process command or socket result immediately

mxinden:
> The `frame` future might be _ready_ with an `Error` from the underlying
socket (i.e. here `libp2p-websocket`). Though given that the result of the
`control_command` `Future` is handled first, `on_control_command` is called
despite `frame` having returned an `Error`. `on_control_command` itself may try
to write to the underlying socket, which will panic given that the socket
returned an error earlier via the `frame` `Future`.

With this patch, once any of `next_stream_command`, `next_control_command` or
`next_frame` `Future` is ready, the result is processed right away, without
additionally polling the remaining pending `Future`s, thus surfacing errors as
early as possible.

See https://github.com/libp2p/rust-libp2p/issues/2598 for details.

@elenaf9 @thomaseizinger could either of you do me a favor and take a look here? Happy to expand on the pull request description in case more context is needed.

### DIFF
--- a/src/connection.rs
+++ b/src/connection.rs
@@ -105,11 +105,7 @@ use futures::{
     stream::{Fuse, FusedStream},
 };
 use nohash_hasher::IntMap;
-use std::{
-    fmt, io,
-    sync::Arc,
-    task::{Context, Poll},
-};
+use std::{fmt, io, sync::Arc, task::Poll};
 
 pub use control::Control;
 pub use stream::{Packet, State, Stream};
@@ -427,7 +423,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
 
             let mut num_terminated = 0;
 
-            let mut next_frame = if self.socket.is_terminated() {
+            let next_frame = if self.socket.is_terminated() {
                 num_terminated += 1;
                 Either::Left(future::pending())
             } else {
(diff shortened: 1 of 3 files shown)


mxinden:
I am not happy with `next` and `next_stream`. I don't think this patch makes things worse. I don't think it improves anything beyond the bug fix.

At this point I don't think we should refactor this Yamux implementation. Instead I suggest to focus on the two new transport implementations https://github.com/libp2p/rust-libp2p/pull/2289 and https://github.com/libp2p/rust-libp2p/pull/2622, as these leverage more powerful stream multiplexing protocols (i.e. QUIC's multiplexer and SCTP).

src/connection.rs
```diff
@@ -477,29 +473,19 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
...
+            match combined_future.await {
+                Either::Left((Either::Left((cmd, _)), _)) => self.on_stream_command(cmd).await?,
+                Either::Left((Either::Right((cmd, _)), _)) => self.on_control_command(cmd).await?,
+                Either::Right((frame, _)) => {
+                    if let Some(stream) =
+                        self.on_frame(frame.transpose().map_err(Into::into)).await?
+                    {
+                        return Ok(Some(stream));
+                    }
                 }
```
thomaseizinger, line 489: Have you considered using the `futures::select!` macro instead of nesting `future::select` calls?

thomaseizinger (approved):
This makes sense to me. Just one suggestion :)

src/connection.rs
mxinden: I have, though I favor `futures::future::select` over `futures::select!` due to the following reasons:

- While more verbose, I find `futures::future::select` easier to reason about than `futures::select!`.
- In general I prefer functions over macros.
- I understand the inner workings of `futures::future::select`. [The implementation](https://docs.rs/futures-util/0.3.21/src/futures_util/future/select.rs.html#101-113) is very simple.
- I don't fully understand the guarantees of `futures::select!`, e.g.:
    > If multiple futures are ready, one will be pseudo-randomly selected at runtime.
    
    https://docs.rs/futures/latest/futures/macro.select.html
    
    What happens to the result of the non-selected ready futures?

elenaf9 (approved):

mxinden:
Been tested without further panics https://github.com/libp2p/rust-libp2p/issues/2598#issuecomment-1206092676

mxinden:
[Tagged](https://github.com/libp2p/rust-yamux/releases/tag/v0.10.2) and [published](https://crates.io/crates/yamux/0.10.2).

PR #148, opened 2022-12-01
fix: Remove dead link to README.md

mxinden:
Release fails with the below otherwise:

```
error: failed to read `readme` file for package `yamux v0.11.0 (/home/xxx/rust-yamux/yamux)`
```

### DIFF
--- a/yamux/Cargo.toml
+++ b/yamux/Cargo.toml
@@ -7,7 +7,6 @@ description = "Multiplexer over reliable, ordered connections"
 keywords = ["network", "protocol"]
 categories = ["network-programming"]
 repository = "https://github.com/paritytech/yamux"
-readme = "README.md"
 edition = "2018"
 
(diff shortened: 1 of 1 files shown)


yamux/Cargo.toml
```diff
@@ -7,7 +7,6 @@ description = "Multiplexer over reliable, ordered connections"
 keywords = ["network", "protocol"]
 categories = ["network-programming"]
 repository = "https://github.com/paritytech/yamux"
-readme = "README.md"
```
thomaseizinger, line 10: Should it be "../README.md" then?

mxinden: My bad. Probably. I thought crates.io can infer it from the repository URL but it doesn't. Too late for the v0.11.0 release. @thomaseizinger can you provide a patch for future releases?

PR #156, opened 2023-05-19
fix: avoid race condition between pending frames and closing stream

thomaseizinger:
Currently, we have a `garbage_collect` function that checks whether any of our streams have been dropped. This can cause a race condition where the channel between a `Stream` and the `Connection` still has pending frames for a stream but dropping a stream causes us to already send a `FIN` flag for the stream.

We fix this by maintaining a single channel for each stream. When a stream gets dropped, the `Receiver` becomes disconnected. We use this information to queue the correct frame (`FIN` vs `RST`) into the buffer. At this point, all previous frames have already been processed and the race condition is thus not present.

Additionally, this also allows us to implement `Stream::poll_flush` by forwarding to the underlying `Sender`. Note that at present day, this only checks whether there is _space_ in the channel, not whether the items have been emitted by the `Receiver`.

We have a PR upstream that might fix this: https://github.com/rust-lang/futures-rs/pull/2746

Fixes: #117.


yamux/src/connection.rs
```diff
@@ -359,6 +362,13 @@ struct Active<T> {
 pub(crate) enum StreamCommand {
     /// A new frame should be sent to the remote.
     SendFrame(Frame<Either<Data, WindowUpdate>>),
+    Flush {
+        id: StreamId,
+        /// How many frames we've queued for sending at the time the flush was requested.
+        num_frames: u64,
```
thomaseizinger, line 368: Wondering: Do we need to send this? We could also just send the waker and simply compare the counters in `Shared`. That will be done anyway by the  implementation of `poll_flush` so we might as well use it in the connection task.

I think that makes this a bit easier to understand.

thomaseizinger: Doubling down on this: I think this means we can contain the state entirely within `Shared` and only store a single `Waker` for flushing.

thomaseizinger: I tried for a bit but can't seem to make a solution that doesn't include this flush mark.

mxinden:
Would the following alternative approach as well be an option?

- Currently we have a single channel for all streams to send `StreamCommand` to the connection. How about a unique channel between each stream and its connection?
- Flushing a stream is simply the matter of calling `Sink::poll_flush` on the `mpsc::Sender`.
- Garbage collection can use the `Poll::Ready(None)` returned by a `mpsc::Receiver` from a stream as a signal instead of `Arc::strong_count`.

Given that the above would not require a new command variant and would not require sending `Waker`s over `mpsc::channel`s, I consider it to be simpler.

@thomaseizinger what do you think? Am I missing something?

thomaseizinger:
> I like that idea! Much simpler. Will implement! :)

There is one caveat. `poll_flush` on `Sender` only checks whether we can send more items into the channel, not whether they have been taken out on the `Receiver` end. Thus, we need to set the capacity to `0` for this to work.

mxinden:
> > I like that idea! Much simpler. Will implement! :)
> 
> There is one caveat. `poll_flush` on `Sender` only checks whether we can send more items into the channel, not whether they have been taken out on the `Receiver` end. Thus, we need to set the capacity to `0` for this to work.
> 
> See [rust-lang/futures-rs#2504](https://github.com/rust-lang/futures-rs/issues/2504).

As far as I can tell, this should be fine. We could even not call `poll_flush` at all. The receiver will still receive all messages and eventually receive `Poll::Ready(None)`. See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=def93d1497874d26f93dabf4a74b3feb.

thomaseizinger:
> > > I like that idea! Much simpler. Will implement! :)
> > 
> > 
> > There is one caveat. `poll_flush` on `Sender` only checks whether we can send more items into the channel, not whether they have been taken out on the `Receiver` end. Thus, we need to set the capacity to `0` for this to work.
> > See [rust-lang/futures-rs#2504](https://github.com/rust-lang/futures-rs/issues/2504).
> 
> As far as I can tell, this should be fine. We could even not call `poll_flush` at all. The receiver will still receive all messages and eventually receive `Poll::Ready(None)`. See [play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=def93d1497874d26f93dabf4a74b3feb](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=def93d1497874d26f93dabf4a74b3feb).

I am not sure what you are trying to say? The point of this PR is to have `poll_flush` park the task if the channel is not yet empty. How can we achieve that if we don't forward to the `poll_flush` call in `Sender`?

FWIW: I attempted to fix this upstream: https://github.com/rust-lang/futures-rs/pull/2746

thomaseizinger:
> * Garbage collection can use the `Poll::Ready(None)` returned by a `mpsc::Receiver` from a stream as a signal instead of `Arc::strong_count`.

Ah I forgot about this.

thomaseizinger:
@mxinden This is now available for review again. I implemented your suggestion. It is the much better design :)

yamux/src/connection.rs
```diff
@@ -490,6 +501,32 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         }
     }
 
+    fn poll_stream_receivers(&mut self, cx: &mut Context) -> Poll<StreamCommand> {
+        for i in (0..self.stream_receivers.len()).rev() {
+            let (id, mut receiver) = self.stream_receivers.swap_remove(i);
```
mxinden, line 506: Polling each stream `Receiver` even though only one might be `Poll::Ready` seems wasteful. Would `SelectAll` not be an option? One would wrap each `Receiver` such that it returns the `StreamId` when the `Receiver` returns `Poll::Ready(None)`. With that `StreamId` we can then call `self.on_drop_stream(id)`.

(On consecutive polls the wrapper can return `Poll::Ready(None)` and thus it would be cleaned up by the `SelectAll`.)

yamux/src/connection.rs
```diff
@@ -821,6 +909,18 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         Action::None
     }
 
+    fn make_new_stream(&mut self, id: StreamId, window: u32, credit: u32) -> Stream {
+        let config = self.config.clone();
+
+        let (sender, receiver) = mpsc::channel(10);
```
mxinden, line 915: I am assuming that this number is chosen arbitrarily, correct? If so, mind documenting that fact? Would allow us to tune it in the future without wondering why `10` was chosen.

yamux/src/connection/stream.rs
```diff
@@ -358,8 +354,10 @@ impl AsyncWrite for Stream {
         Poll::Ready(Ok(n))
     }
 
-    fn poll_flush(self: Pin<&mut Self>, _: &mut Context) -> Poll<io::Result<()>> {
-        Poll::Ready(Ok(()))
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<io::Result<()>> {
+        self.sender
+            .poll_flush_unpin(cx)
+            .map_err(|_| self.write_zero_err())
```
mxinden, line 360: :tada: neat!

yamux/src/connection.rs
```diff
@@ -541,6 +573,71 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
+                // The stream was dropped without calling `poll_close`.
+                // We have already received a FIN from remote and send one
+                // back which closes the stream for good.
+                State::RecvClosed => {
+                    let mut header = Header::data(stream_id, 0);
+                    header.fin();
+                    Some(Frame::new(header))
+                }
+                // The stream was properly closed. We either already have
+                // or will at some later point send our FIN frame.
```
mxinden, line 600:
```suggestion
                // The stream was properly closed. We either already sent our FIN frame.
```

Given that we only call `poll_stream_receivers` when the socket is flushed, I think it is safe to assume here that the `FIN` is sent, right?

mxinden (commented):
This looks good to me overall, though one suggestion around using a `SelectAll`.

yamux/src/connection.rs
thomaseizinger: I had a `SelectAll` based design before but dropped it because I couldn't detect `None` from the outside.

Detecting that requires another data structure for translating between the stream commands and the actual frames in connection (the wrapping you mentioned).

This also needs to work for the `Cleanup` and `Close` case.

I tried hiding it in a custom collection object but we need access to the stream `IntMap` upon drop so that needs even more refactoring but would be a clean design.

I can invest the time if you want but I don't think it is a quick 30min refactoring.

thomaseizinger: I was wrong, it was actually a super quick refactoring, now that I've seen how tokio's `StreamMap` works. They recommend an additional wrapper around `Stream`. The trick is to wrap the output of the `Stream` again such that you get a tuple of `(key, Option<Item>)` which allows you to detect closing of the stream.

yamux/Cargo.toml
```diff
@@ -16,6 +16,7 @@ nohash-hasher = "0.2"
 parking_lot = "0.12"
 rand = "0.8.3"
 static_assertions = "1"
+pin-project = "1.1.0"
```
mxinden, line 19: Instead of using `pin-project`, we could as well implement `Unpin` for `Stream`.

``` rust
impl Unpin for Stream {}
```

That said, I doubt there is any async Rust code-base without `pin-project` in its dependency tree already. Thus fine to include it here as well.

mxinden (approved):
Thanks for the follow-up. Also great to be aware of [`StreamMap`](https://dtantsur.github.io/rust-openstack/tokio/stream/struct.StreamMap.html#method.remove).

Issue #157, opened 2023-05-19
Remove `ControlledConnection` and everything around it


thomaseizinger:
Either after or with https://github.com/libp2p/rust-yamux/issues/154, we should also remove the old `ControlledConnection` as I consider it the legacy interface now that we have a first-class poll-based API.

PR #158, opened 2023-05-22
refactor: move more tests to poll-API based interface

thomaseizinger:
Whilst debugging some test failures for #156, I rewrote one more test to use the poll-based interface.

Related: #157.


mxinden (approved):
Thanks for the clean-up!

PR #165, opened 2023-07-03
Allow to configure the maximum size for the frame channel between one connection and one of its stream.

stormshield-damiend:
While doing some benchmarks using iperf3 on a network client / server application using Yamux + Tokio TcpStream + rustls we noticed that the mpsc between a Connection and one of its Stream was filling too fast leading to lower bandwidth.

With the default hard-coded value (10 packets) with the client and server application running inside two dockers on the same laptop, we measured 315 Mbits/sec. With an increased value of 16*1024 packets, we measured 426 Mbits/sec.

This PR adds an API that allow setting this value in YamuxConfig and leave the default value as is.

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -876,7 +876,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     fn make_new_stream(&mut self, id: StreamId, window: u32, credit: u32) -> Stream {
         let config = self.config.clone();
 
-        let (sender, receiver) = mpsc::channel(10); // 10 is an arbitrary number.
+        let (sender, receiver) = mpsc::channel(config.max_stream_backlog);
         self.stream_receivers.push(TaggedStream::new(id, receiver));
         if let Some(waker) = self.no_streams_waker.take() {
             waker.wake();
--- a/yamux/src/lib.rs
+++ b/yamux/src/lib.rs
@@ -50,6 +50,9 @@ pub type Result<T> = std::result::Result<T, ConnectionError>;
 /// actual upper bound is this value + number of clones.
 const MAX_COMMAND_BACKLOG: usize = 32;
 
+/// Arbitrary limit of our per stream channels.
+const MAX_STREAM_BACKLOG: usize = 10;
+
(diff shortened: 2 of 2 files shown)


thomaseizinger:
I wonder if we need to make this configurable or can't just increase the size of the channel. As documented, 10 was picked completely arbitrary.

A question around your benchmarks: Did you run the stream and the connection on different tasks? I'd imagine the performance to be quite bad if the same task has to switch back and forth between transferring data on the stream (which writes to the queue) vs writing data to the wire (which empties this queue).

The former one is actually the default in `rust-libp2p` if you poll streams as part of a `ConnectionHandler` and not in a separately spawned task.

How big each individual frame can be also depends on how much credit the remote gave you. See https://github.com/libp2p/rust-yamux/issues/162 for a related discussion. Thus, I'd expect that because of our fairly small default window size, you need a lot of space in this buffer _initially_ to achieve good performance. As soon as the receive window has been increased, one initial frame could easily be several MB big at which point the performance should be quite good, despite a small buffer size.

I am a bit hesitant to just expose this as a configuration option. I'd rather like to explore if we can figure out what a good value is. Given that this is only for outgoing frames, I think we don't need to be worried about DoS attacks and can just set this to a high value.

stormshield-damiend:
Hi Thomas,

Some more context.

Here are the Yamux settings we use
```
let mut yamux_config = YamuxConfig::default();
yamux_config.set_window_update_mode(WindowUpdateMode::on_read());
yamux_config.set_receive_window_size(16 * 1024 * 1024);
yamux_config.set_max_buffer_size(16 * 1024 * 1024);
yamux_config.set_max_stream_backlog(16 * 1024);
```
We use Yamux to stream some data between client and server, on a dedicated stream we have a pattern like send length  of data (2 bytes) + send data (around 1400).

We use tokio runtime in current_thread mode. I can run the same tests in multi_thread mode if needed.
We have many spawns all over our code to write and poll data.

About your proposal to expose or not this setting, it is true we could find a good value based on the initial size of the max_buffer_size.

stormshield-damiend:
Hi Thomas,

sure i can run some more tests using different size, it may take some time before i get back with the results

PR #167, opened 2023-07-06
fix: garbage collect `Stream` when dropped

### LINKED ISSUE #166: Dropped `Stream`s are never removed until connection closes?
I'm integrating yamux into my app, which creates many new streams on a single connection. I am finding that when I run my hacked-together benchmark tool, new streams on the server side are continually accepted and never removed, even though the `yamux::Stream` objects used in my app are properly being dropped, until the `max_num_streams` limit in yamux is hit and my benchmark consequently dies.

In looking at `Active<T>` in `connection.rs`, it seems that the `max_num_streams` limit is enforced based on the length of `self.streams`. The only time streams are removed from that map are:
1) in `on_drop_stream`
2) in `drop_all_streams`, when the connection is dropped

Setting aside (2), which works fine, I am having trouble seeing how (1) is supposed to work? It looks like `on_drop_stream` is called in `Active<T>::poll` when the corresponding channel in `self.stream_receivers` is closed (i.e. receiver returns `None`). However, the clone of the `Stream` that's stored in `self.streams` itself includes a clone of the `mpsc::Sender<StreamCommand>`. Even when my app drops its `Stream` (and thus its `Sender`), the clone that's stored in the `Active<T>` persists, which looks like a circular reference that prevents that `Stream` from ever being removed until the whole connection is dropped.

I feel like I must be missing something? How is this supposed to work, so that streams can be removed (and thus not count towards the max streams limit) before the connection is taken down?

aschran:
This fixes an issue where a `Stream` would never be removed until its connection is closed.

Resolves #166.


yamux/src/connection.rs
```diff
@@ -540,12 +543,12 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_drop_stream(&mut self, id: StreamId) {
-        let stream = self.streams.remove(&id).expect("stream not found");
+        let s = self.streams.remove(&id).expect("stream not found");
 
-        log::trace!("{}: removing dropped {}", self.id, stream);
-        let stream_id = stream.id();
+        log::trace!("{}: removing dropped", self.id);
```
mxinden, line 548:
```suggestion
        log::trace!("{}: removing dropped stream {}", self.id, id);
```

Would make the log message more useful.

yamux/src/connection.rs
```diff
@@ -540,12 +543,12 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_drop_stream(&mut self, id: StreamId) {
```
mxinden, line 545:
```suggestion
    fn on_drop_stream(&mut self, stream_id: StreamId) {
```

Would be a bit easier to read and remove the need for `let stream_id = id;` below. I know, not directly related, thus feel free to ignore.

yamux/src/connection.rs
```diff
@@ -350,7 +353,7 @@ struct Active<T> {
     socket: Fuse<frame::Io<T>>,
     next_id: u32,
 
-    streams: IntMap<StreamId, Stream>,
+    streams: IntMap<StreamId, Arc<Mutex<Shared>>>,
```
mxinden, line 356:
```suggestion
    streams: IntMap<StreamId, Arc<Mutex<stream::Shared>>>,
```

Adds more context to the name `Shared`, i.e. makes it clear that it is shared with the stream, and not e.g. with other connections.

mxinden (commented):
Very well done @aschran. Thank you for reporting, debugging and fixing. I appreciate your work here.

I missed this in my review in https://github.com/libp2p/rust-yamux/pull/156#issuecomment-1556858738.

Overall this looks good to me. Missing steps:

- I would like to give @thomaseizinger a chance to review as well.
- @aschran would you mind adding a changelog entry to `CHANGELOG.md`?

Again, well done. Thank you.

mxinden:
This likely warrants a patch release (i.e. `v0.11.1`). I can get that done once we merged here.

mxinden:
@aschran @thomaseizinger I merged latest `master` and added a test that fails on `master` without this patch.

aschran:
Committed your suggestions. Thanks for the review and confirmation + test!

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
   Users have to move to the `poll_` functions of `Connection`.
   See [PR #164](https://github.com/libp2p/rust-yamux/pull/164).
 
+- Fix a bug where `Stream`s would not be dropped until their corresponding
+  `Connection` was dropped.
```
thomaseizinger, line 8:
```suggestion
- Fix a bug where `Stream`s would not be dropped until their corresponding `Connection` was dropped.
  See [PR #167](https://github.com/libp2p/rust-yamux/pull/167).
```

yamux/src/connection.rs
```diff
@@ -539,13 +542,13 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
 
-    fn on_drop_stream(&mut self, id: StreamId) {
-        let stream = self.streams.remove(&id).expect("stream not found");
+    fn on_drop_stream(&mut self, stream_id: StreamId) {
+        let s = self.streams.remove(&id).expect("stream not found");
 
-        log::trace!("{}: removing dropped {}", self.id, stream);
-        let stream_id = stream.id();
+        log::trace!("{}: removing dropped stream {}", self.id, id);
+        let stream_id = id;
```
thomaseizinger, line 549: This doesn't compile.

I'd suggest we just name it `stream_id` for the entire function.

yamux/src/connection.rs
```diff
@@ -103,13 +103,16 @@ use closing::Closing;
...
 use nohash_hasher::IntMap;
+use parking_lot::Mutex;
 use std::collections::VecDeque;
 use std::task::{Context, Waker};
 use std::{fmt, sync::Arc, task::Poll};
 
 use crate::tagged_stream::TaggedStream;
 pub use stream::{Packet, State, Stream};
 
+use self::stream::Shared;
```
thomaseizinger, line 114: Can we express this as a `crate::` import and merge it with the others above?

yamux/src/connection.rs
```diff
@@ -350,7 +353,7 @@ struct Active<T> {
     socket: Fuse<frame::Io<T>>,
     next_id: u32,
 
-    streams: IntMap<StreamId, Stream>,
+    streams: IntMap<StreamId, Arc<Mutex<stream::Shared>>>,
```
thomaseizinger, line 356: I wonder if it would turn out cleaner if `Shared` would hold `Arc<Mutex>` internally? That would remove the boilerplate around locking and some noise from the type signatures.

thomaseizinger (approved):
Thank you!

Ironically, the fact that all of `Stream` was shared bugged me already in the past because it is conceptually hard to understand. The fact that it now actually caused a bug annoys me that I didn't follow my gut feeling there :)

I made one suggestion on how I think we could make this a bit cleaner. Let me know if you are up for it, otherwise happy to merge this as is once CI passes!

yamux/src/connection.rs
aschran: Actually the import is no longer needed after using qualified `stream::Shared`, removed

yamux/src/connection.rs
aschran: Would prefer to merge the fix and leave this for future work if that's ok with you, since it looks like a bit more involved of a change

thomaseizinger (approved):

thomaseizinger:
Can you look at the CI failures? Otherwise LGTM!

aschran:
> Can you look at the CI failures? Otherwise LGTM!

I'll admit I'm not sure what's up with those. If I run `cargo build` on this branch locally it works fine for me. The errors in CI suggest it's not running on the same code I'm looking at? `yamux/src/connection.rs` is only 927 lines long and yet errors are showing lines 942-943? It doesn't seem I have access to trigger a re-run of CI myself.

mxinden:
Our CI first merges the target branch, here `master`, and then executes the tests. 

We recently merged #153. While git does not identify it as a merge conflict, it does conflict with this pull request.

I pushed a merge resolving the non-git-merge-conflict conflicts. @thomaseizinger this moves the `is_outbound` method inline and duplicates the `is_pending_ack` on `Shared`.

mxinden (approved):
Merging here.

@thomaseizinger let me know in case you would like to see any follow-ups given the latest changes.

@aschran once more, thank you for the work. Let me know in case you would like me to backport this change to `v0.11`.

aschran:
@mxinden yes, if you have time to backport that would be great! Appreciate the diligent review from you both.

thomaseizinger:
> @thomaseizinger let me know in case you would like to see any follow-ups given the latest changes.

None that are critical for a release.

mxinden:
@aschran released happened through https://github.com/libp2p/rust-yamux/pull/169. Let me know if that is good enough for you.

aschran:
That works, thanks for letting me know!

PR #168, opened 2023-07-19
feat: don't split header and body across TCP packets

stormshield-pj50:
Currently, header and body of a frame are written to the underlying IO source in two steps. Depending on how eager the underlying stream is about flushing data, this can lead to two TCP packets being sent instead of header and body ending up in one.

By refactoring the internal implementation of `Frame`, we can reference it as a single byte slice and as a result, only use a single call to `poll_write` without additional allocations per frame.

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -880,7 +880,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         if stream_id == CONNECTION_ID || self.streams.contains_key(&stream_id) {
             let mut hdr = Header::ping(frame.header().nonce());
             hdr.ack();
-            return Action::Ping(Frame::no_body(hdr));
+            return Action::Ping(Frame::new(hdr));
         }
         log::trace!(
             "{}/{}: ping for unknown stream, possibly dropped earlier: {:?}",
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -880,7 +880,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         if stream_id == CONNECTION_ID || self.streams.contains_key(&stream_id) {
             let mut hdr = Header::ping(frame.header().nonce());
             hdr.ack();
-            return Action::Ping(Frame::new(hdr));
+            return Action::Ping(Frame::no_body(hdr));
         }
         log::trace!(
             "{}/{}: ping for unknown stream, possibly dropped earlier: {:?}",
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -91,7 +91,6 @@ mod cleanup;
 mod closing;
 mod stream;
 
-use crate::frame::header::HEADER_SIZE;
 use crate::tagged_stream::TaggedStream;
 use crate::{
     error::ConnectionError,
@@ -356,7 +355,7 @@ struct Active<T> {
     socket: Fuse<frame::Io<T>>,
     next_id: u32,
 
-    streams: IntMap<u32, Arc<Mutex<stream::Shared>>>,
+    streams: IntMap<StreamId, Arc<Mutex<stream::Shared>>>,
     stream_receivers: SelectAll<TaggedStream<StreamId, mpsc::Receiver<StreamCommand>>>,
     no_streams_waker: Option<Waker>,
 
@@ -520,14 +519,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
 
         if extra_credit > 0 {
             let mut frame = Frame::window_update(id, extra_credit);
-            let mut parsed_frame = frame.parse_mut().expect("valid frame");
-            parsed_frame.header_mut().syn();
-            log::trace!(
-                "{}/{}: sending initial {}",
-                self.id,
-                id,
-                parsed_frame.header()
-            );
+            frame.header_mut().syn();
+            log::trace!("{}/{}: sending initial {}", self.id, id, frame.header());
             self.pending_frames.push_back(frame.into());
         }
 
@@ -538,18 +531,17 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         }
 
         log::debug!("{}: new outbound {} of {}", self.id, stream, self);
-        self.streams.insert(id.val(), stream.clone_shared());
+        self.streams.insert(id, stream.clone_shared());
 
         Poll::Ready(Ok(stream))
     }
 
     fn on_send_frame(&mut self, frame: Frame<Either<Data, WindowUpdate>>) {
-        let parsed_frame = frame.parse().expect("valid frame");
         log::trace!(
             "{}/{}: sending: {}",
             self.id,
-            parsed_frame.stream_id(),
-            parsed_frame.header()
+            frame.header().stream_id(),
+            frame.header()
         );
         self.pending_frames.push_back(frame.into());
     }
@@ -561,10 +553,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_drop_stream(&mut self, stream_id: StreamId) {
-        let s = self
-            .streams
-            .remove(&stream_id.val())
-            .expect("stream not found");
+        let s = self.streams.remove(&stream_id).expect("stream not found");
 
         log::trace!("{}: removing dropped stream {}", self.id, stream_id);
         let frame = {
@@ -575,7 +564,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 State::Open { .. } => {
                     let mut header = Header::data(stream_id, 0);
                     header.rst();
-                    Some(Frame::from_header(header))
+                    Some(Frame::new(header))
                 }
                 // The stream was dropped without calling `poll_close`.
                 // We have already received a FIN from remote and send one
@@ -583,7 +572,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 State::RecvClosed => {
                     let mut header = Header::data(stream_id, 0);
                     header.fin();
-                    Some(Frame::from_header(header))
+                    Some(Frame::new(header))
                 }
                 // The stream was properly closed. We already sent our FIN frame.
                 // The remote may be out of credit though and blocked on
@@ -596,7 +585,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                         // which we will never send, so reset the stream now.
                         let mut header = Header::data(stream_id, 0);
                         header.rst();
-                        Some(Frame::from_header(header))
+                        Some(Frame::new(header))
                     } else {
                         // The remote has either still credit or will be given more
                         // (due to an enqueued window update or because the update
@@ -620,7 +609,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             frame
         };
         if let Some(f) = frame {
-            log::trace!("{}/{}: sending: {:?}", self.id, stream_id, f);
+            log::trace!("{}/{}: sending: {}", self.id, stream_id, f.header());
             self.pending_frames.push_back(f.into());
         }
     }
@@ -632,12 +621,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     /// Otherwise we process the frame and potentially return a new `Stream`
     /// if one was opened by the remote.
     fn on_frame(&mut self, frame: Frame<()>) -> Result<Option<Stream>> {
-        let parsed_frame = frame.parse().expect("valid frame");
-        log::trace!("{}: received: {:?}", self.id, parsed_frame);
+        log::trace!("{}: received: {}", self.id, frame.header());
 
-        if parsed_frame.has_flag(header::ACK) {
-            let id = parsed_frame.stream_id();
-            if let Some(stream) = self.streams.get(&id.val()) {
+        if frame.header().flags().contains(header::ACK) {
+            let id = frame.header().stream_id();
+            if let Some(stream) = self.streams.get(&id) {
                 stream
                     .lock()
                     .update_state(self.id, id, State::Open { acknowledged: true });
@@ -647,7 +635,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             }
         }
 
-        let action = match parsed_frame.tag().expect("valid header's tag") {
+        let action = match frame.header().tag() {
             Tag::Data => self.on_data(frame.into_data()),
             Tag::WindowUpdate => self.on_window_update(&frame.into_window_update()),
             Tag::Ping => self.on_ping(&frame.into_ping()),
@@ -658,25 +646,21 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             Action::New(stream, update) => {
                 log::trace!("{}: new inbound {} of {}", self.id, stream, self);
                 if let Some(f) = update {
-                    let pf = f.parse().expect("valid frame");
-                    log::trace!("{}/{}: sending update", self.id, pf.stream_id());
+                    log::trace!("{}/{}: sending update", self.id, f.header().stream_id());
                     self.pending_frames.push_back(f.into());
                 }
                 return Ok(Some(stream));
             }
             Action::Update(f) => {
-                let pf = f.parse().expect("valid frame");
-                log::trace!("{}: sending update: {:?}", self.id, pf.header());
+                log::trace!("{}: sending update: {:?}", self.id, f.header());
                 self.pending_frames.push_back(f.into());
             }
             Action::Ping(f) => {
-                let pf = f.parse().expect("valid frame");
-                log::trace!("{}/{}: pong", self.id, pf.stream_id());
+                log::trace!("{}/{}: pong", self.id, f.header().stream_id());
                 self.pending_frames.push_back(f.into());
             }
             Action::Reset(f) => {
-                let pf = f.parse().expect("valid frame");
-                log::trace!("{}/{}: sending reset", self.id, pf.stream_id());
+                log::trace!("{}/{}: sending reset", self.id, f.header().stream_id());
                 self.pending_frames.push_back(f.into());
             }
             Action::Terminate(f) => {
@@ -689,12 +673,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_data(&mut self, frame: Frame<Data>) -> Action {
-        let parsed_frame = frame.parse().expect("valid frame");
-        let stream_id = parsed_frame.stream_id();
+        let stream_id = frame.header().stream_id();
 
-        if parsed_frame.has_flag(header::RST) {
+        if frame.header().flags().contains(header::RST) {
             // stream reset
-            if let Some(s) = self.streams.get_mut(&stream_id.val()) {
+            if let Some(s) = self.streams.get_mut(&stream_id) {
                 let mut shared = s.lock();
                 shared.update_state(self.id, stream_id, State::Closed);
                 if let Some(w) = shared.reader.take() {
@@ -707,15 +690,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             return Action::None;
         }
 
-        let is_finish = parsed_frame.has_flag(header::FIN); // half-close
+        let is_finish = frame.header().flags().contains(header::FIN); // half-close
 
-        if parsed_frame.has_flag(header::SYN) {
+        if frame.header().flags().contains(header::SYN) {
             // new stream
             if !self.is_valid_remote_id(stream_id, Tag::Data) {
                 log::error!("{}: invalid stream id {}", self.id, stream_id);
                 return Action::Terminate(Frame::protocol_error());
             }
-            if parsed_frame.body().len() > DEFAULT_CREDIT as usize {
+            if frame.body().len() > DEFAULT_CREDIT as usize {
                 log::error!(
                     "{}/{}: 1st body of stream exceeds default credit",
                     self.id,
@@ -723,7 +706,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 );
                 return Action::Terminate(Frame::protocol_error());
             }
-            if self.streams.contains_key(&stream_id.val()) {
+            if self.streams.contains_key(&stream_id) {
                 log::error!("{}/{}: stream already exists", self.id, stream_id);
                 return Action::Terminate(Frame::protocol_error());
             }
@@ -738,16 +721,14 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 if is_finish {
                     shared.update_state(self.id, stream_id, State::RecvClosed);
                 }
-                shared.window = shared.window.saturating_sub(parsed_frame.body_len());
-                shared.buffer.push(frame.into_buffer(), HEADER_SIZE);
+                shared.window = shared.window.saturating_sub(frame.body_len());
+                shared.buffer.push(frame.into_body());
 
                 if matches!(self.config.window_update_mode, WindowUpdateMode::OnReceive) {
                     if let Some(credit) = shared.next_window_update() {
                         shared.window += credit;
-
                         let mut frame = Frame::window_update(stream_id, credit);
-                        let mut parsed_frame = frame.parse_mut().expect("valid frame");
-                        parsed_frame.header_mut().ack();
+                        frame.header_mut().ack();
                         window_update = Some(frame)
                     }
                 }
@@ -755,13 +736,13 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             if window_update.is_none() {
                 stream.set_flag(stream::Flag::Ack)
             }
-            self.streams.insert(stream_id.val(), stream.clone_shared());
+            self.streams.insert(stream_id, stream.clone_shared());
             return Action::New(stream, window_update);
         }
 
-        if let Some(s) = self.streams.get_mut(&stream_id.val()) {
+        if let Some(s) = self.streams.get_mut(&stream_id) {
             let mut shared = s.lock();
-            if parsed_frame.body().len() > shared.window as usize {
+            if frame.body().len() > shared.window as usize {
                 log::error!(
                     "{}/{}: frame body larger than window of stream",
                     self.id,
@@ -781,10 +762,10 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 );
                 let mut header = Header::data(stream_id, 0);
                 header.rst();
-                return Action::Reset(Frame::from_header(header));
+                return Action::Reset(Frame::new(header));
             }
-            shared.window = shared.window.saturating_sub(parsed_frame.body_len());
-            shared.buffer.push(frame.into_buffer(), HEADER_SIZE);
+            shared.window = shared.window.saturating_sub(frame.body_len());
+            shared.buffer.push(frame.into_body());
             if let Some(w) = shared.reader.take() {
                 w.wake()
             }
@@ -815,12 +796,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_window_update(&mut self, frame: &Frame<WindowUpdate>) -> Action {
-        let parsed_frame = frame.parse().expect("valid frame");
-        let stream_id = parsed_frame.stream_id();
+        let stream_id = frame.header().stream_id();
 
-        if parsed_frame.has_flag(header::RST) {
+        if frame.header().flags().contains(header::RST) {
             // stream reset
-            if let Some(s) = self.streams.get_mut(&stream_id.val()) {
+            if let Some(s) = self.streams.get_mut(&stream_id) {
                 let mut shared = s.lock();
                 shared.update_state(self.id, stream_id, State::Closed);
                 if let Some(w) = shared.reader.take() {
@@ -833,15 +813,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             return Action::None;
         }
 
-        let is_finish = parsed_frame.has_flag(header::FIN); // half-close
+        let is_finish = frame.header().flags().contains(header::FIN); // half-close
 
-        if parsed_frame.has_flag(header::SYN) {
+        if frame.header().flags().contains(header::SYN) {
             // new stream
             if !self.is_valid_remote_id(stream_id, Tag::WindowUpdate) {
                 log::error!("{}: invalid stream id {}", self.id, stream_id);
                 return Action::Terminate(Frame::protocol_error());
             }
-            if self.streams.contains_key(&stream_id.val()) {
+            if self.streams.contains_key(&stream_id) {
                 log::error!("{}/{}: stream already exists", self.id, stream_id);
                 return Action::Terminate(Frame::protocol_error());
             }
@@ -850,7 +830,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 return Action::Terminate(Frame::protocol_error());
             }
 
-            let credit = parsed_frame.credit() + DEFAULT_CREDIT;
+            let credit = frame.header().credit() + DEFAULT_CREDIT;
             let mut stream = self.make_new_inbound_stream(stream_id, credit);
             stream.set_flag(stream::Flag::Ack);
 
@@ -859,13 +839,13 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                     .shared()
                     .update_state(self.id, stream_id, State::RecvClosed);
             }
-            self.streams.insert(stream_id.val(), stream.clone_shared());
+            self.streams.insert(stream_id, stream.clone_shared());
             return Action::New(stream, None);
         }
 
-        if let Some(s) = self.streams.get_mut(&stream_id.val()) {
+        if let Some(s) = self.streams.get_mut(&stream_id) {
             let mut shared = s.lock();
-            shared.credit += parsed_frame.credit();
+            shared.credit += frame.header().credit();
             if is_finish {
                 shared.update_state(self.id, stream_id, State::RecvClosed);
             }
@@ -892,16 +872,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     }
 
     fn on_ping(&mut self, frame: &Frame<Ping>) -> Action {
-        let parsed_frame = frame.parse().expect("valid frame");
-        let stream_id = parsed_frame.stream_id();
-        if parsed_frame.has_flag(header::ACK) {
+        let stream_id = frame.header().stream_id();
+        if frame.header().flags().contains(header::ACK) {
             // pong
             return Action::None;
         }
-        if stream_id == CONNECTION_ID || self.streams.contains_key(&stream_id.val()) {
-            let mut hdr = Header::ping(parsed_frame.nonce());
+        if stream_id == CONNECTION_ID || self.streams.contains_key(&stream_id) {
+            let mut hdr = Header::ping(frame.header().nonce());
             hdr.ack();
-            return Action::Ping(Frame::from_header(hdr));
+            return Action::Ping(Frame::new(hdr));
         }
         log::trace!(
             "{}/{}: ping for unknown stream, possibly dropped earlier: {:?}",
@@ -968,8 +947,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             // - Its ID is odd and we are the client.
             // - Its ID is even and we are the server.
             .filter(|(id, _)| match self.mode {
-                Mode::Client => StreamId::new(**id).is_client(),
-                Mode::Server => StreamId::new(**id).is_server(),
+                Mode::Client => id.is_client(),
+                Mode::Server => id.is_server(),
             })
             .filter(|(_, s)| s.lock().is_pending_ack())
             .count()
@@ -992,7 +971,7 @@ impl<T> Active<T> {
     fn drop_all_streams(&mut self) {
         for (id, s) in self.streams.drain() {
             let mut shared = s.lock();
-            shared.update_state(self.id, StreamId::new(id), State::Closed);
+            shared.update_state(self.id, id, State::Closed);
             if let Some(w) = shared.reader.take() {
                 w.wake()
             }
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -500,7 +500,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         }
     }
 
-    fn poll_new_outbound(&mut self, cx: &mut Context<'_>) -> Poll<Result<Stream>> {
+    fn poll_new_outbound(&mut self, cx: &Context<'_>) -> Poll<Result<Stream>> {
         if self.streams.len() >= self.config.max_num_streams {
             log::error!("{}: maximum number of streams reached", self.id);
             return Poll::Ready(Err(ConnectionError::TooManyStreams));
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -91,6 +91,7 @@ mod cleanup;
 mod closing;
 mod stream;
 
+use crate::frame::header::HEADER_SIZE;
 use crate::tagged_stream::TaggedStream;
 use crate::{
     error::ConnectionError,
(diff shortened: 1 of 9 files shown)


thomaseizinger:
Thank you for digging into this! This looks like a great improvement. Is there a reason why this needs to be configurable? We will always need to write a header so can't we just always reserve 12 bytes for the header?

thomaseizinger:
> concatenate header and body and try to write them as one byte blob?

This would mean an additional allocation per frame which is not great. But, we could introduce a new constructor for `Frame<Data>` that gets initialized from a slice. Currently, `AsyncWrite::poll_write` for `Stream` allocates anyway for each `Frame`. If we move this allocation into the `Frame` type, we can directly allocate an additional 12 bytes and store the header alongside it.

For this to work, we'd also have to change how `Io` accesses the bytes to be written but for all other frame-types, we don't have any payload, meaning we need something like "encode frame into byte vec" and use that within `Io::start_send`.

Thoughts?

stormshield-pj50:
> Thank you for digging into this! This looks like a great improvement. Is there a reason why this needs to be configurable? We will always need to write a header so can't we just always reserve 12 bytes for it?

You're right we can assume that if we have enough space it must be 12 bytes, but in that case we have to expose header::HEADER_SIZE to the outside.

thomaseizinger:
> > Thank you for digging into this! This looks like a great improvement. Is there a reason why this needs to be configurable? We will always need to write a header so can't we just always reserve 12 bytes for it?
> 
> You're right we can assume that if we have enough space it must be 12 bytes, but in that case we have to expose header::HEADER_SIZE to the outside.

I am not sure I fully understand what you mean?

Regardless, I think that https://github.com/libp2p/rust-yamux/pull/168#issuecomment-1642365611 is an avenue worth exploring because it should avoid the need for any of these APIs by merging header + body into one chunk for data frames which should always result in one TCP packet, assuming it fits.

stormshield-pj50:
> > concatenate header and body and try to write them as one byte blob?
> 
> This would mean an additional allocation per frame which is not great. But, we could introduce a new constructor for `Frame<Data>` that gets initialized from a slice. Currently, `AsyncWrite::poll_write` for `Stream` allocates anyway for each `Frame`. If we move this allocation into the `Frame` type, we can directly allocate an additional 12 bytes and store the header alongside it.
> 
> For this to work, we'd also have to change how `Io` accesses the bytes to be written but for all other frame-types, we don't have any payload, meaning we need something like "encode frame into byte vec" and use that within `Io::start_send`.
> 
> Thoughts?

So you propose to allocate directly into a new Frame<Data> constructor: allocate 12 bytes more for the header, encode the header into the buffer then copy the slice into the buffer at offset 12. Eventually it means the Stream API set_reserved_header_size() becomes useless, isn't it ?

thomaseizinger:
> > > concatenate header and body and try to write them as one byte blob?
> > 
> > 
> > This would mean an additional allocation per frame which is not great. But, we could introduce a new constructor for `Frame<Data>` that gets initialized from a slice. Currently, `AsyncWrite::poll_write` for `Stream` allocates anyway for each `Frame`. If we move this allocation into the `Frame` type, we can directly allocate an additional 12 bytes and store the header alongside it.
> > For this to work, we'd also have to change how `Io` accesses the bytes to be written but for all other frame-types, we don't have any payload, meaning we need something like "encode frame into byte vec" and use that within `Io::start_send`.
> > Thoughts?
> 
> So you propose to allocate directly into a new Frame constructor: allocate 12 bytes more for the header, encode the header into the buffer then copy the slice into the buffer at offset 12. Eventually it means the Stream API set_reserved_header_size() becomes useless, isn't it ?

Correct! `Frame<Data>` is the only variant that actually has a body so I think it makes more sense to embed this feature there. We probably need to slightly change the internals of `Frame` and the interaction within `Io` where we extract the data to be written to the socket.

I'd have to look into the details but off the top of my head I am thinking:

- Expose a `&[u8]` from `Frame`
- For `Frame<Data>`, we just expose the internal buffer (where header and body are concatenated)
- For all other frames, we just expose the slice of the buffer
- `Io` needs to keep track of how many bytes of the current frame it already wrote. We do something like this already I believe.

The last point makes sure we don't allocate again in `Io` but instead request the slice and index into it based on how many bytes we already wrote to the socket.

stormshield-pj50:
> > > > concatenate header and body and try to write them as one byte blob?
> > > 
> > > 
> > > This would mean an additional allocation per frame which is not great. But, we could introduce a new constructor for `Frame<Data>` that gets initialized from a slice. Currently, `AsyncWrite::poll_write` for `Stream` allocates anyway for each `Frame`. If we move this allocation into the `Frame` type, we can directly allocate an additional 12 bytes and store the header alongside it.
> > > For this to work, we'd also have to change how `Io` accesses the bytes to be written but for all other frame-types, we don't have any payload, meaning we need something like "encode frame into byte vec" and use that within `Io::start_send`.
> > > Thoughts?
> > 
> > 
> > So you propose to allocate directly into a new Frame constructor: allocate 12 bytes more for the header, encode the header into the buffer then copy the slice into the buffer at offset 12. Eventually it means the Stream API set_reserved_header_size() becomes useless, isn't it ?
> 
> Correct! `Frame<Data>` is the only variant that actually has a body so I think it makes more sense to embed this feature there. We probably need to slightly change the internals of `Frame` and the interaction within `Io` where we extract the data to be written to the socket.
> 
> I'd have to look into the details but off the top of my head I am thinking:
> 
>     * Expose a `&[u8]` from `Frame`
> 
>     * For `Frame<Data>`, we just expose the internal buffer (where header and body are concatenated)
> 
>     * For all other frames, we just expose the slice of the buffer
> 
>     * `Io` needs to keep track of how many bytes of the current frame it already wrote. We do something like this already I believe.
> 
> 
> The last point makes sure we don't allocate again in `Io` but instead request the slice and index into it based on how many bytes we already wrote to the socket.

OK that's makes sense. I will try to work on it. In the same time I was wondering if we could avoid the copy of the buffer in poll_write() (or in Frame<Data> as you suggest) by implementing a second API on Stream similar to the Sink trait. For instance a start_send() method that takes ownership of a BytesMut and indicates whether there is some room at the start of the buffer for a Yamux header. With this we could avoid an allocation and a copy. What do you think about it ?

yamux/src/frame.rs
```diff
@@ -18,18 +18,28 @@ use std::{convert::TryInto, num::TryFromIntError};
...
+use self::header::HEADER_SIZE;
+
 /// A Yamux message frame consisting of header and body.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Eq)]
 pub struct Frame<T> {
     header: Header<T>,
-    body: Vec<u8>,
+    buf: Vec<u8>,
+}
```
thomaseizinger, line 28: How about turning `Frame` into an enum that is either just a header or a Vec?

For data frames, we know that the first 12 bytes are the header so we could theoretically re-parse them if needed.

But really, all we need is a function to return a slice, right?

I'd like to make this inner representation as unambiguous as possible.

stormshield-pj50: > How about turning `Frame` into an enum that is either just a header or a Vec?
> 
> For data frames, we know that the first 12 bytes are the header so we could theoretically re-parse them if needed.
> 
> But really, all we need is a function to return a slice, right?
> 
> I'd like to make this inner representation as unambiguous as possible.

```
enum Frame<T> {
   Header(Header<T>),
   HeaderBody(Vec<u8>)
}
```

I think it's a little strange to have a Header variant holding a struct Header while the HeaderBody variant direclty holds the header (plus body) in binary. The patch I proposed yesterday makes Frame<T> holds both the header and a buffer with always 12 bytes at its beginning for binary encoding the header.

thomaseizinger: > I think it's a little strange to have a Header variant holding a struct Header while the HeaderBody variant direclty holds the header (plus body) in binary.

The idea would be that we are never storing redundant data. If I understand your patch correctly, we are redundantly storing `Header` in the case of a data frame, right?

Instead of thinking about it as `Header` and `HeaderBody`, we can think of it as `Header` and `Encoded`, maybe that makes more sense?

stormshield-pj50: > > I think it's a little strange to have a Header variant holding a struct Header while the HeaderBody variant direclty holds the header (plus body) in binary.
> 
> The idea would be that we are never storing redundant data. If I understand your patch correctly, we are redundantly storing `Header` in the case of a data frame, right?

Yes that's it, except that we are not storing Header redundantly, we use the 12 first bytes of `buf` field Frame<T> to binary encode the header when needed in io.rs and send it on the wire. That's why the Frame<T> PartialEq implementation skips the comparison of the 12 first bytes of buf Frame<T> field.
> 
> Instead of thinking about it as `Header` and `HeaderBody`, we can think of it as `Header` and `Encoded`, maybe that makes more sense?
Why not have the `Header` variant binary encoded ? Isn't it useless to keep it under a `Header` struct ?

thomaseizinger: But this means we now allocate also for header-only frames, right? Which technically is not necessary but maybe it doesn't matter.

Why even keep the header around then and not always encode it directly into a buf?

mxinden:
Thank you @stormshield-pj50 for posting your findings here!

> Our software has some gains of 50% throughput with this patch, as we are mostly sending small buffers with Yamux.

Can you add more details on the setup where you are seeing a 50% throughput increase? What is the latency and bandwidth between the two endpoints? Is it a LAN, WAN or routed over the Internet? How many TCP connections (or other flows) run concurrently? What is the lifetime of a stream on a single connection?

stormshield-pj50:
> Thank you @stormshield-pj50 for posting your findings here!
> 
> > Our software has some gains of 50% throughput with this patch, as we are mostly sending small buffers with Yamux.
> 
> Can you add more details on the setup where you are seeing a 50% throughput increase? What is the latency and bandwidth between the two endpoints? Is it a LAN, WAN or routed over the Internet? How many TCP connections (or other flows) run concurrently? What is the lifetime of a stream on a single connection?

Hi Max,
Sorry for the late reply, I was on vacation !
Using libp2p + Yamux + rustls, we tunnel a single iperf3 TCP connection between two local docker containers. As we prefix each tunneled packet with its size in the Yamux stream, in the end we write 4 packets for one tunneled packet. The Yamux stream lives as long as the connection.

yamux/src/frame.rs
stormshield-pj50: > But this means we now allocate also for header-only frames, right? Which technically is not necessary but maybe it doesn't matter.

Yes you're right we allocate 12 unneeded bytes for header-only frames, but I think it's unimportant.

> 
> 
> Why even keep the header around then and not always encode it directly into a buf?

So you propose to have:
```
pub struct Frame {
  buf: Vec<u8>
}
```
And always encode the header at the beginning of buf. But In that case how do you deal with the current Frame<T> type state machinery (Frame<()>, Frame<Data>, ...) ?

thomaseizinger: > > But this means we now allocate also for header-only frames, right? Which technically is not necessary but maybe it doesn't matter.
> 
> Yes you're right we allocate 12 unneeded bytes for header-only frames, but I think it's unimportant.

I am not too worried about an additional allocation itself but I don't like the redundant storing of data in the current solution.

> > Why even keep the header around then and not always encode it directly into a buf?
> 
> So you propose to have:
> 
> ```
> pub struct Frame {
>   buf: Vec<u8>
> }
> ```
> 
> And always encode the header at the beginning of buf. But In that case how do you deal with the current Frame type state machinery (Frame<()>, Frame< Data >, ...) ?

I haven't thought about that yet. Maybe it can be removed altogether? I am open to proposals :)

stormshield-pj50: Maybe we could use the Google's zerocopy crate in order to parse and serialize automatically the Yamux header to/from bytes. With this we only keep into contiguous memory the slice bytes (header + data) and can read/write header fields transparently. And of course, the Frame<T> type state continues working.

I just pushed a quick poc [there](https://github.com/stormshield-pj50/yamux_poc_zerocopy/blob/master/src/main.rs).  The only drawback is that we have to store the tag field as an u8 into the Header struct, as zerocopy can't derive FromBytes on enums with less variants than it's underlying representation type size. However it seems they are working on TryFromBytes and we could use it in the future.

thomaseizinger: This looks good to me! The PoC doesn't go all the way in regards to type-safety though because one can update the tag without updating the type parameter. We'd have to keep the current conversion methods and set the tag implicitly.

stormshield-pj50: I have updated all the code of Yamux with the zerocopy header parsing, the patch is quite straightforward.
I'm still using a Frame<T> in Yamux data structs, but now Frame<T> only holds a single buffer (frame + data).
The Frame<T> is zero-copied-parsed into a ParsedFrame<&[u8], T> or ParsedFrame<&mut [u8], T> when needed. The ParsedFrame struct holds a zerocopy Ref on the header's bytes and a slice on the frame's body.
Please also note that Chunks::push() now takes an offset into the vector, and that the underlying Cursor is built with it. That prevents to remove the 12 header bytes from the frame's vector (and avoid some memory copies).

yamux/src/frame.rs
```diff
@@ -13,89 +13,144 @@ mod io;
...
+        Self::make_parsed_frame(header, body)
+    }
+
+    pub fn into_buffer(self) -> Vec<u8> {
+        self.buffer
+    }
+
+    pub fn buffer(&self) -> &[u8] {
+        &self.buffer
+    }
```
thomaseizinger, line 86: Can we move these further down such that Git shows them as a diff to the previous `body()` and `into_body` functions?

yamux/src/connection/stream.rs
```diff
@@ -218,7 +218,8 @@ impl Stream {
             drop(shared);
 
             let mut frame = Frame::window_update(self.id, credit).right();
-            self.add_flag(frame.header_mut());
+            let mut parsed_frame = frame.parse_mut().expect("valid frame");
+            self.add_flag(parsed_frame.header_mut());
```
thomaseizinger, line 222: Can we make this more ergonomic somehow? It is duplicated with line 376.

yamux/src/connection/stream.rs
```diff
@@ -367,18 +369,19 @@ impl AsyncWrite for Stream {
...
+        let mut parsed_frame = frame.parse_mut().expect("valid frame");
+        self.add_flag(parsed_frame.header_mut());
         log::trace!("{}/{}: write {} bytes", self.conn, self.id, n);
 
         // technically, the frame hasn't been sent yet on the wire but from the perspective of this data structure, we've queued the frame for sending
         // We are tracking this information:
         // a) to be consistent with outbound streams
         // b) to correctly test our behaviour around timing of when ACKs are sent. See `ack_timing.rs` test.
-        if frame.header().flags().contains(ACK) {
+        if parsed_frame.header().flags().contains(ACK) {
```
thomaseizinger, line 384: Parsing the frame is cheap I'd assume? Can we make this more ergonomic by adding a `has_flag` function to `Frame` so we don't need to keep around the `parsed_frame` here?

yamux/src/frame.rs
```diff
@@ -13,89 +13,144 @@ mod io;
...
+/// A Yamux message frame consisting of header and body in a single buffer
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Frame<T> {
-    header: Header<T>,
-    body: Vec<u8>,
+    buffer: Vec<u8>,
+    _marker: std::marker::PhantomData<T>,
+}
+
+impl<T> Default for Frame<T> {
```
thomaseizinger, line 31: I don't think this should exist. Doesn't this more or less circumvent all the type safety around `_marker` if one can instantiation a `Frame` of any `T` just like that?

yamux/src/frame.rs
```diff
@@ -105,52 +160,68 @@ impl Frame<Data> {
...
+    pub fn window_update(id: StreamId, credit: u32) -> Frame<WindowUpdate> {
+        Frame::from_header(Header::window_update(id, credit))
     }
+}
 
-    pub fn body_len(&self) -> u32 {
-        // Safe cast since we construct `Frame::<Data>`s only with
-        // `Vec<u8>` of length [0, u32::MAX] in `Frame::data` above.
-        self.body().len() as u32
+impl Frame<GoAway> {
```
thomaseizinger, line 173: Can we re-order the code such that this doesn't show up as this much diff?

yamux/src/connection.rs
```diff
@@ -609,7 +620,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             frame
         };
         if let Some(f) = frame {
-            log::trace!("{}/{}: sending: {}", self.id, stream_id, f.header());
+            let pf = f.parse().expect("valid frame");
```
thomaseizinger, line 623: It is quite unfortunate that we have all these panic paths now. I guess "parsing" using zerocopy is very cheap? Can we validate that a frame is valid when we initially construct it and re-parse it every time we need it in order to avoid this?

thomaseizinger (commented):
Thank you!

Couple of questions regarding ergonomics of these new APIs.

yamux/src/frame.rs
stormshield-pj50: ok

yamux/src/connection/stream.rs
stormshield-pj50: ok add_flag() now takes a Frame and parses it and update the frame's flags

yamux/src/connection/stream.rs
stormshield-pj50: add_flag() now returns the flags, so no need to call contains again

yamux/src/frame.rs
stormshield-pj50: True. I previously added Default in order to call std::mem::take on a Frame in io.rs. Now I have changed std::mem::take with std::mem::replace.

yamux/src/frame.rs
stormshield-pj50: ok

mxinden (changes requested):
I am not yet convinced that this is a feature worth supporting. Keep in mind, I might be missing something.

> Using libp2p + Yamux + rustls, we tunnel a single iperf3 TCP connection between two local docker containers.

Do I understand correctly, that this is localhost networking only? Is this your primary use-case? If not, can you please provide numbers on a more realistic network environment (i.e. an environment with significant latency and bandwidth limitations)?

In my eyes this pull request introduces significant complexity. At this point I don't think this complexity is worth the improvement of a minor use-case (localhost networking) of this library.

thomaseizinger:
> In my eyes this pull request introduces significant complexity. At this point I don't think this complexity is worth the improvement of a minor use-case (localhost networking) of this library.

Is it really "significant"? I think we should be able to implement this such that it is almost all contained within `Frame`.

I think it makes perfect sense and is quite intuitive that a frame can be encoded into a single byte buffer. Not doing that is the weird thing, no?

I agree that we may need to polish the API a bit more but as long as it is contained, I am happy to accept this.

stormshield-pj50:
> I am not yet convinced that this is a feature worth supporting. Keep in mind, I might be missing something.
> 
> > Using libp2p + Yamux + rustls, we tunnel a single iperf3 TCP connection between two local docker containers.
> 
> Do I understand correctly, that this is localhost networking only? Is this your primary use-case? If not, can you please provide numbers on a more realistic network environment (i.e. an environment with significant latency and bandwidth limitations)?

The first aim of the dev is to prevent sending two TCP packets on the wire each time some data is written to the Yamux stream (one packet for Yamux header and one packet for data). Thus we can save one useless standalone header packet and at least 14 (ETH) + 20 (IP) +  20 (TCP)  = 54 bytes for each bunch of data written to the stream. Take a look at the start of the PR for a packet capture that show this. That also means less network syscalls and a more scalable code.

Our local test is just for benchmarking purpose, however I think with or without latency the important thing is to limit the number of packets sent on the wire.

> 
> In my eyes this pull request introduces significant complexity. At this point I don't think this complexity is worth the improvement of a minor use-case (localhost networking) of this library.

thomaseizinger:
@stormshield-pj50 I'll have a play with this API next week, ideally, I'd like to encapsulate more of this in the `Frame` type. Currently the zero-parsing stuff leaks to many other components which isn't very nice and makes things harder to understand.

Do you think we can refactor this such that it is completely an implementation detail of `Frame`?

thomaseizinger:
> @stormshield-pj50 I'll have a play with this API next week, ideally, I'd like to encapsulate more of this in the `Frame` type. Currently the zero-parsing stuff leaks to many other components which isn't very nice and makes things harder to understand.
> 
> Do you think we can refactor this such that it is completely an implementation detail of `Frame`?

I think, if we can contain this optimisation mostly in `Frame`, then we have a lot less complexity to deal with and can maybe convince @mxinden to land this :)

thomaseizinger:
@stormshield-pj50 I pushed a commit that outlines a different approach for using `zerocopy`. It needs more work but I hope that it demonstrates, how I'd like to contain the functionality within the `Frame` module. As you can see from the diff, `connection.rs` and other modules are pretty much unaffected.

Unfortunately, `zerocopy` doesn't handle enums very well as you've probably encountered yourself in the previous commits. They seem to be working on it but I am not sure when they'll land that work. For now, I think the workaround isn't too bad.

Let me know if you are willing to continue down this path :)

yamux/src/frame.rs
```diff
@@ -13,89 +13,130 @@ mod io;
...
+    buffer: Vec<u8>,
+    _marker: PhantomData<T>,
 }
 
 impl<T> Frame<T> {
-    pub fn new(header: Header<T>) -> Self {
-        Frame {
-            header,
-            body: Vec::new(),
+    pub(crate) fn new(header: Header<T>) -> Self {
```
thomaseizinger, line 34: FIXME: Remove `(crate)` to minimise diff.

yamux/src/frame.rs
```diff
@@ -13,89 +13,130 @@ mod io;
...
 pub use io::FrameDecodeError;
 pub(crate) use io::Io;
 
-/// A Yamux message frame consisting of header and body.
-#[derive(Clone, Debug, PartialEq, Eq)]
+use crate::HeaderDecodeError;
+
+use self::header::{Flags, Tag, HEADER_SIZE};
+
+/// A Yamux message frame consisting of header and body in a single buffer
```
thomaseizinger, line 26: FIXME: Wrong now.

yamux/src/frame.rs
```diff
@@ -13,89 +13,130 @@ mod io;
...
+
+        Self {
+            buffer,
+            _marker: PhantomData,
         }
     }
 
-    pub fn header(&self) -> &Header<T> {
-        &self.header
+    pub(crate) fn header(&self) -> &Header<T> {
```
thomaseizinger, line 48: FIXME: Reduce diff here too by removing `(crate)`.

yamux/src/frame.rs
```diff
@@ -107,50 +148,24 @@ impl Frame<Data> {
...
-
-    pub fn body_len(&self) -> u32 {
-        // Safe cast since we construct `Frame::<Data>`s only with
-        // `Vec<u8>` of length [0, u32::MAX] in `Frame::data` above.
-        self.body().len() as u32
-    }
-
-    pub fn into_body(self) -> Vec<u8> {
-        self.body
-    }
```
thomaseizinger, line 123: FIXME: Order functions such that this diff doesn't show up.

yamux/src/frame.rs
```diff
@@ -107,50 +148,24 @@ impl Frame<Data> {
...
     }
 }
 
 impl Frame<GoAway> {
-    pub fn term() -> Self {
-        Frame {
-            header: Header::term(),
-            body: Vec::new(),
-        }
+    pub fn term() -> Frame<GoAway> {
```
thomaseizinger, line 160: FIXME: Reduce diff:

```suggestion
    pub fn term() -> Self {
```

yamux/src/frame/io.rs
```diff
@@ -174,19 +125,18 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Stream for Io<T> {
         let mut this = &mut *self;
         loop {
             log::trace!("{}: read: {:?}", this.id, this.read_state);
-            match this.read_state {
+            let state = mem::replace(&mut this.read_state, ReadState::Init);
```
thomaseizinger, line 128: FIXME: Inline this variable.

thomaseizinger (commented):
Left some more notes on what still needs work!

stormshield-pj50:
> @stormshield-pj50 I pushed a commit that outlines a different approach for using `zerocopy`. It needs more work but I hope that it demonstrates, how I'd like to contain the functionality within the `Frame` module. As you can see from the diff, `connection.rs` and other modules are pretty much unaffected.
> 
> Unfortunately, `zerocopy` doesn't handle enums very well as you've probably encountered yourself in the previous commits. They seem to be working on it but I am not sure when they'll land that work. For now, I think the workaround isn't too bad.

Yes zerocopy poor enum handling is a pity. But their TryFromBytes wip could make the job in theory.
> 
> Let me know if you are willing to continue down this path :)

Containing zerocopy into the Frame module is a good idea! However I won't be able to work on it this week, maybe next week. I'll finish the work and push a new commit. Thank you for your help!

thomaseizinger:
> > Let me know if you are willing to continue down this path :)
> 
> Containing zerocopy into the Frame module is a good idea! However I won't be able to work on it this week, maybe next week. I'll finish the work and push a new commit. Thank you for your help!

No rush from my end! Let me know if you have any questions on the code!

thomaseizinger:
Note: Having this would be helpful in debugging the TCP RTT issue in https://github.com/libp2p/test-plans/pull/304. At the moment, header and body are always split into two TCP packets which makes this rather cumbersome to follow along!

yamux/src/connection.rs
```diff
@@ -738,30 +721,28 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 if is_finish {
                     shared.update_state(self.id, stream_id, State::RecvClosed);
                 }
-                shared.window = shared.window.saturating_sub(parsed_frame.body_len());
-                shared.buffer.push(frame.into_buffer(), HEADER_SIZE);
```
stormshield-pj50: I introduced the second parameter to avoid the copy of the body. Do you really want to revert this optimization ? I did notice you suggest to use BytesMut in order to split of the body, however I'm not sure its a good idea to do it in this PR.

yamux/src/frame.rs
```diff
@@ -27,17 +27,59 @@ use self::header::{Flags, Tag, HEADER_SIZE};
...
 pub struct Frame<T> {
     buffer: Vec<u8>,
-    _marker: std::marker::PhantomData<T>,
+    _marker: PhantomData<T>,
 }
 
 impl<T> Frame<T> {
-    pub(crate) fn new(buffer: Vec<u8>) -> Self {
+    pub(crate) fn new(header: Header<T>) -> Self {
+        let total_buffer_size = HEADER_SIZE + header.len().val() as usize;
```
stormshield-pj50, line 35: You can't allocate space for body here, it depends on the kind of header, for instance Header<WindowUpdate> encodes credit into its len field without providing a body.

yamux/src/frame/io.rs
stormshield-pj50, line 128: mem::replace will leave read_state in ReadState::Init state if we exit the function with one of the two ready! macro below.

yamux/src/connection.rs
thomaseizinger: I actually re-did the entire patch-set to reduce as much diff as possible. Compared to `master`, there is no regression here :)

I'd suggest we do one thing at a time and focus on optimising how we write frames with this PR. Reducing allocations can be left to a later point!

thomaseizinger:
@stormshield-pj50 Thanks for the comments! I noticed that you commented on the commits directly instead of the entire PR diff. I'd suggest to not discuss intermediary changes and instead work based on the effective diff to `master`!

yamux/src/frame/io.rs
thomaseizinger: Good catch!

thomaseizinger:
I spent some more cycles on this and I think this is now a quite clean implementation. Would appreciate another review @mxinden! The implementation is now entirely contained within `Frame` so there isn't much complexity and it is all guarded by the type-system, in some places even more than before!

thomaseizinger:
@stormshield-pj50 Can you do me two final favours?

1. Test this in your system to make sure it has the desired effect.
2. Create a pull-request against `rust-libp2p` so we run our interoperability tests and check that we didn't break anything.

yamux/src/frame/io.rs
```diff
@@ -152,66 +114,58 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Sink<Frame<()>> for Io<T> {
...
                 ReadState::Header {
                     ref mut offset,
                     ref mut buffer,
                 } => {
                     if *offset == header::HEADER_SIZE {
-                        let header = match header::decode(buffer) {
-                            Ok(hd) => hd,
-                            Err(e) => return Poll::Ready(Some(Err(e.into()))),
-                        };
+                        let frame = Frame::try_from_header_buffer(buffer)?;
```
stormshield-pj50, line 148: If someone send data frames with a large len field, Frame::try_from_header_buffer will call Frame::new and allocate large buffers for nothing because they will be dropped by the check on max_body_len. That's why I did introduce ensure_buffer_len() to resize the buffer after the check and prevent a potential denial of service attack.

yamux/src/frame/header.rs
```diff
@@ -323,65 +348,47 @@ impl fmt::Display for StreamId {
...
-    buf[1] = hdr.tag as u8;
-    buf[2..4].copy_from_slice(&hdr.flags.0.to_be_bytes());
-    buf[4..8].copy_from_slice(&hdr.stream_id.0.to_be_bytes());
-    buf[8..HEADER_SIZE].copy_from_slice(&hdr.length.0.to_be_bytes());
-    buf
-}
-
-/// Decode a [`Header`] value.
+// Decode a [`Header`] value.
 pub fn decode(buf: &[u8; HEADER_SIZE]) -> Result<Header<()>, HeaderDecodeError> {
```
stormshield-pj50, line 381: I'm not convinced that the decoding should be done on the raw buffer, as we have introduced zerocopy and can work instead on the struct header.

stormshield-pj50:
> @stormshield-pj50 Can you do me two final favours?
> 
>     1. Test this in your system to make sure it has the desired effect.
> 
>     2. Create a pull-request against `rust-libp2p` so we run our interoperability tests and check that we didn't break anything.

Sure! I will do it today.

stormshield-pj50:
> > @stormshield-pj50 Can you do me two final favours?
> > ```
> > 1. Test this in your system to make sure it has the desired effect.
> > 
> > 2. Create a pull-request against `rust-libp2p` so we run our interoperability tests and check that we didn't break anything.
> > ```
> 
> Sure! I will do it today.

Integration tests PR in rust-libp2p: https://github.com/libp2p/rust-libp2p/pull/4598

yamux/src/frame/io.rs
thomaseizinger: Good catch, we should probably just do the check inside then! I'll send a follow-up commit.

yamux/src/frame/header.rs
thomaseizinger: It isn't really good use of the type system to allow construction of types that are in an invalid state. Thus, until `zerocopy` can give us these guarantees, I'd rather have a dedicated `decode` function. This somewhat ensures that we are not working with a `Header` that has invalid values set.

stormshield-pj50:
> > @stormshield-pj50 Can you do me two final favours?
> > ```
> > 1. Test this in your system to make sure it has the desired effect.
> > 
> > 2. Create a pull-request against `rust-libp2p` so we run our interoperability tests and check that we didn't break anything.
> > ```
> 
> Sure! I will do it today.

On our side, our CI tests are all OK.

thomaseizinger (approved):
Good from my end but also needs @mxinden's approval.

mxinden (commented):
Overall I am fine moving forward here. The rational behind the change makes sense. Thank you for the continuous work.

While intuitively I expect this to have no performance drawback, I would still like to see some benchmarks. Would you mind:

- Testing the latest version in your setup and reporting the latest numbers?
- Running the `concurrent.rs` benchmarks, especially the `unconstrained`, comparing this patch with latest `master`, e.g. using `critcmp`?
- Ideally (optional) plug this into https://github.com/libp2p/test-plans/blob/master/perf/README.md#libp2p-performance-benchmarking.

yamux/src/connection.rs
```diff
@@ -500,7 +500,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         }
     }
 
-    fn poll_new_outbound(&mut self, cx: &mut Context<'_>) -> Poll<Result<Stream>> {
+    fn poll_new_outbound(&mut self, cx: &Context<'_>) -> Poll<Result<Stream>> {
```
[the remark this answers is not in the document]
thomaseizinger: Triggered by a clippy lint, we can put it back but given that it is an internal API, I wouldn't bother.

yamux/src/frame/header.rs
```diff
@@ -10,18 +10,35 @@
...
 use futures::future::Either;
 use std::fmt;
+use std::hash::{Hash, Hasher};
+use zerocopy::big_endian::{U16, U32};
+use zerocopy::{AsBytes, FromBytes, FromZeroes};
 
 /// The message frame header.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq, FromBytes, AsBytes, FromZeroes)]
+#[repr(packed)]
```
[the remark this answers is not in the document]
thomaseizinger: This is necessary for `zerocopy` to do its job so I don't think we have a choice unless we want to trade `zerocopy` for more allocations.

thomaseizinger:
> * Running the `concurrent.rs` benchmarks, especially the `unconstrained`, comparing this patch with latest `master`, e.g. using `critcmp`?

I'd take these with a grain of salt. When I worked on the refactoring to `poll` APIs, I couldn't get really consistent numbers out of repeated benchmark runs **without** code changes, i.e. it was hard to establish a base-line. Plugging it into your perf-setup might make more sense.

thomaseizinger:
Great results! Looks like we actually measurably improved performance :)

stormshield-pj50:
Results from our iperf3 CI test:
```
With Yamux master:
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   324 MBytes   272 Mbits/sec   26             sender
[  5]   0.00-10.05  sec   322 MBytes   269 Mbits/sec                  receiver

With Yamux PR:
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec   831 MBytes   697 Mbits/sec  32              sender
[  5]   0.00-10.04  sec   829 MBytes   693 Mbits/sec                  receiver
```

stormshield-pj50:
> We just discussed this in the [open maintainers call](https://github.com/libp2p/rust-libp2p/discussions/4606) and @umgefahren made a really good suggestion: Instead of doing all this work with `zerocopy`, we could just use [`AsyncWrite::poll_write_vectored`](https://docs.rs/futures/0.3.28/futures/io/trait.AsyncWrite.html#method.poll_write_vectored). That would allow us to pass two buffers down to the underlying implementation.
> 
> It is unfortunate that this idea only comes up now but I think it is well worth exploring as it would avoid the complexity of `zerocopy`.

You can't be sure that poll_write_vectored() is well implemented on the underlying stream, as the default implementation takes the first non empty buffer and call poll_write() on it.  I tested it a couple of weeks ago before submitting this PR and it was working with a TcpStream but not with a TlsStream on top of it.

thomaseizinger:
> > We just discussed this in the [open maintainers call](https://github.com/libp2p/rust-libp2p/discussions/4606) and @umgefahren made a really good suggestion: Instead of doing all this work with `zerocopy`, we could just use [`AsyncWrite::poll_write_vectored`](https://docs.rs/futures/0.3.28/futures/io/trait.AsyncWrite.html#method.poll_write_vectored). That would allow us to pass two buffers down to the underlying implementation.
> > It is unfortunate that this idea only comes up now but I think it is well worth exploring as it would avoid the complexity of `zerocopy`.
> 
> You can't be sure that poll_write_vectored() is well implemented on the underlying stream, as the default implementation takes the first non empty buffer and call poll_write() on it. I tested it a couple of weeks ago before submitting this PR and it was working with a TcpStream but not with a TlsStream on top of it.

But isn't that just a shortcoming of the `TlsStream` then? Shouldn't we patch those impls instead of making this more complex?

thomaseizinger:
Which TLS library are you using?

stormshield-pj50:
> Which TLS library are you using?

tokio-rustls

thomaseizinger:
> > > > We just discussed this in the [open maintainers call](https://github.com/libp2p/rust-libp2p/discussions/4606) and @umgefahren made a really good suggestion: Instead of doing all this work with `zerocopy`, we could just use [`AsyncWrite::poll_write_vectored`](https://docs.rs/futures/0.3.28/futures/io/trait.AsyncWrite.html#method.poll_write_vectored). That would allow us to pass two buffers down to the underlying implementation.
> > > > It is unfortunate that this idea only comes up now but I think it is well worth exploring as it would avoid the complexity of `zerocopy`.
> > > 
> > > 
> > > You can't be sure that poll_write_vectored() is well implemented on the underlying stream, as the default implementation takes the first non empty buffer and call poll_write() on it. I tested it a couple of weeks ago before submitting this PR and it was working with a TcpStream but not with a TlsStream on top of it.
> > 
> > 
> > But isn't that just a shortcoming of the `TlsStream` then? Shouldn't we patch those impls instead of making this more complex?
> 
> On our side we won't spend more work time on this PR. We think the implementation is good enough:
> 
>  * zerocopy is a mainstream crate designed by Google and used by many Fuchsia OS devices.
>  * internally zerocopy is just a safe cast with endianness bytes swap when needed.
>  * the PR solution is self contained and we think it's not a good idea to rely on an underlying stream implementation of poll_write_vectored(), which by the way is optional in the specifications.
>  * we have been using this code for months in our code base and it's working fine with improved performance as shown before in the previous discussion.

Thank you for writing down your view-points. `zerocopy` is definitely a well-designed library and I am not disputing its production-readiness. I do want to stress though that it isn't intrinsically necessary for this feature.

If we didn't care about reduced allocations, we could easily keep the `Frame` type as is and simply create a new buffer and write that in a single pass to the underlying stream. The only thing that `zerocopy` gives us is reducing this allocation. At the same time, we currently have to implement hacks because `zerocopy` doesn't allow for any validations during parsing.

Unfortunately, despite being familiar with `AsyncWrite`, it completely slipped my mind that we could use `poll_write_vectored` to achieve the exact same thing without taking on another dependency and deviating from what would be the "natural" design of `Frame` when ignoring performance optimisations.

I cannot follow your reasoning why we shouldn't rely on `AsyncWrite::poll_write_vectored`. In fact, I find using that is a very clean solution. The abstraction is specifically there to optimise performance when it is possible to write multiple buffers at the same time. `tokio-rustls` doesn't override that in its [`AsyncWrite`](https://docs.rs/tokio-rustls/latest/src/tokio_rustls/lib.rs.html#522-553) implementation but I would be very surprised if they wouldn't accept a patch to do that.

Overall, I am hesitant to merge this as is knowing that there is an easier and more maintainable solution. Our time is limited and we need to carefully consider, what we take on as maintenance responsibility.

Thank you for iterating on this solution with me. I wish we would have talked about `AsyncWrite::poll_write_vectored` earlier before going down this pathway. I accept that you don't want to put any more time into it. I hope you understand that we don't want to take on more maintenance burden than necessary.

If I find some free time next week, I might check how hard it would be change to `AsyncWrite::poll_write_vectored`. I have a feeling it would be quite easy!

thomaseizinger:
> `tokio-rustls` doesn't override that in its [`AsyncWrite`](https://docs.rs/tokio-rustls/latest/src/tokio_rustls/lib.rs.html#522-553) implementation but I would be very surprised if they wouldn't accept a patch to do that.

I've opened issue with them to discuss the implementation: https://github.com/rustls/tokio-rustls/issues/26. I think that properly implementing this could be a nice optimization that the entire async rustls ecosystem can benefit from which is another reason on why we shouldn't build a specialised implementation here in `rust-yamux` for this problem.

stormshield-pj50:
> > `tokio-rustls` doesn't override that in its [`AsyncWrite`](https://docs.rs/tokio-rustls/latest/src/tokio_rustls/lib.rs.html#522-553) implementation but I would be very surprised if they wouldn't accept a patch to do that.
> 
> I've opened issue with them to discuss the implementation: [rustls/tokio-rustls#26](https://github.com/rustls/tokio-rustls/issues/26). I think that properly implementing this could be a nice optimization that the entire async rustls ecosystem can benefit from which is another reason on why we shouldn't build a specialised implementation here in `rust-yamux` for this problem.

OK that's fine for us. However if you want to cover all Tokio ecosystem libp2p related, you should also take a look at [Tokio Tunsgtenite WebSockets](https://docs.rs/tokio-tungstenite/latest/tokio_tungstenite/) and also Yamux over Yamux as used by libp2p relay_v2.

thomaseizinger:
I am closing this as stale.

For reference, I started working on correctly forwarding vectored writes but ran out of time. Here are the drafts / issues:

- https://github.com/rustls/tokio-rustls/pull/25
- https://github.com/rustls/tokio-rustls/issues/26

PR #172, opened 2023-08-17
docs: remove mentions of old API components

### LINKED ISSUE #171: Doc for v0.12.0 is outdated
For example, https://docs.rs/yamux/0.12.0/yamux/index.html still refers to the removed `Control` struct.

Gowee:
`Control` and `Connection::next_stream` has been removed since 0.11 and 0.12, in favor of poll-based methods on `Connection`.
  
The PR updated doc comments regarding those structs. Some outdated non-doc comments at the top of connection.rs (L12) are unchanged, because I am not familiar with the crate enough to rewrite that part.

This fixes #171.


mxinden (approved):
Thank you for the help!

PR #176, opened 2023-11-23
feat: auto-tune (dynamic) stream receive window

mxinden:
## Motivation

Yamux provides flow control on the stream level, preventing a fast sender from overwhelming a slow receiver, in other words for the receiver to exerts backpressure to the sender. The receiver does so through communicating a flow control window in-band through `WindowUpdate` messages, basically granting the sender send-credit on a continuous basis.

The big question here is: _how large should the window be_. In other words, how much data should the sender be able to send, before waiting on another `WindowUpdate` from the sender. Choosing the window size too small results in the sender not being able to make use of the entire bandwidth. Choosing the window size too large results in a high memory footprint in the form of a buffer on the receiver side with the size of the window, and delays the backpressure signal from the receiver to the sender.

## Status quo

Today the Rust Yamux window size is fixed. The [default value is 256 KB](https://github.com/libp2p/rust-yamux/blob/1e46505721c66d05ebaffcd8ddd32cd9d541d7b7/yamux/src/lib.rs#L109). Assuming a connection round-trip-time of 60ms, this allows for up to 33 Mbit/s throughput per stream (`256*1024/0,06*8=33.33`). Obviously this is not ideal when running on e.g. a cloud instance with up to 10Gbit/s.

One can bump the default window size, to e.g. 16 MB. But then again, this might be too large for some deployments (low latency and/or low throughput) or too small for others (high latency and/or high throughput).

See https://github.com/libp2p/rust-yamux/issues/162 for a larger discussion on the various options.

newplot(16)

[Single stream throughput on a 60ms TCP connection with up to 5 Gbit/s machine bandwidth.](https://github.com/libp2p/test-plans/blob/master/perf/README.md)

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=9a4c96952368e6432fe51fd12008d31ca30f63ea#branch

## Solution

The solution is to auto-tune the stream receive window, making it dynamic based on the round-trip time and the estimated bandwidth, in other words aiming for the [bandwidth-delay-product](https://en.wikipedia.org/wiki/Bandwidth-delay_product). This is nothing novel. In fact the Linux kernel [auto-tunes the TCP receive buffer](https://linux.die.net/man/7/tcp), [QUIC implementations do it](https://www.rfc-editor.org/rfc/rfc9000.html#name-increasing-flow-control-lim) and the [Go Yamux implementation does it](https://github.com/libp2p/go-yamux/issues/9).

## Implementation

This pull request does the following:

- Send Yamux' Pings on an interval to measure the connection round-trip-time.
- Dynamically grow the stream receive window based on the round-trip-time and the estimated bandwidth. For a detailed analysis of the used algorithm see [here](https://docs.google.com/document/d/1SExkMmGiz8VYzV3s9E35JQlJ73vhzCekKkDi85F1qCE/edit) and [here](https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit). Long story short, we do what QUIC does.

## Result

The results show that this pull request is a game-changer for any Yamux deployments. While low-resourced deployments maintain the benefit of small buffers, high resource deployments eventually end-up with a window of roughly the bandwidth-delay-product (ideal).

newplot(17)

[Same setup. Single stream throughput on a 60ms TCP connection with up to 5 Gbit/s machine bandwidth.](https://github.com/libp2p/test-plans/blob/master/perf/README.md)

https://observablehq.com/@libp2p-workspace/performance-dashboard?branch=9a4c96952368e6432fe51fd12008d31ca30f63ea#branch

(As one can see, there is still room for improvement, i.e. 1.3 Gbit/s is not 5 Gbit/s. Though that might be due to some other parts of our stack. More to come in the future.)

---

Fixes https://github.com/libp2p/rust-yamux/issues/162.

Preliminary performance results: https://github.com/libp2p/test-plans/pull/332/


yamux/src/connection/stream.rs
```diff
@@ -422,29 +609,143 @@ impl AsyncWrite for Stream {
     }
 }
 
+impl Drop for Stream {
+    fn drop(&mut self) {
+        let max_receive_window_size = self.shared.lock().max_receive_window_size;
+        let mut accumulated_max_stream_windows = self.accumulated_max_stream_windows.lock();
+
+        debug_assert!(
+            *accumulated_max_stream_windows >= (max_receive_window_size - DEFAULT_CREDIT) as usize,
```
thomaseizinger, line 618: Extract the RHS part into a variable.

yamux/src/connection/stream.rs
```diff
@@ -422,29 +609,143 @@ impl AsyncWrite for Stream {
...
+                self.accumulated_max_stream_windows.lock().clone(),
+            )),
+            rtt: self.rtt.clone(),
+            last_window_update: self.last_window_update.clone(),
+        }
+    }
+}
+
+#[cfg(test)]
+impl quickcheck::Arbitrary for Stream {
```
thomaseizinger, line 646: This looks like a massive hack for the benefit of calling `next_window_update` in a proptest. It would be better to extract everything around window-updates into a dedicated struct, test that using `proptest` and embed it within `Stream`.

yamux/src/lib.rs
```diff
@@ -95,70 +76,89 @@ pub enum WindowUpdateMode {
...
+    ///
+    /// Set to `None` to disable the per stream maximum receive window.
     ///
     /// # Panics
     ///
     /// If the given receive window is < 256 KiB.
-    pub fn set_receive_window(&mut self, n: u32) -> &mut Self {
-        assert!(n >= DEFAULT_CREDIT);
-        self.receive_window = n;
+    pub fn set_max_stream_receive_window(&mut self, n: Option<u32>) -> &mut Self {
```
thomaseizinger, line 115: Why do we need to make this configurable? Isn't it okay for a single stream to increase its buffer as long as it stays within the configured max memory usage of the entire connection?

yamux/src/lib.rs
```diff
@@ -95,70 +76,89 @@ pub enum WindowUpdateMode {
...
+        self.check();
         self
     }
+
+    // TODO: Consider doing the check on creation, not on each builder method call.
+    fn check(&self) {
+        assert!(self.max_stream_receive_window.unwrap_or(u32::MAX) >= DEFAULT_CREDIT);
+        assert!(
+            self.max_connection_receive_window >= self.max_num_streams * DEFAULT_CREDIT as usize
+        );
```
thomaseizinger, line 160: Definitely replace these with `Result`s.

yamux/src/frame/io.rs
```diff
@@ -20,24 +20,24 @@ use std::{
     task::{Context, Poll},
 };
 
+const MAX_FRAME_BODY_LEN: usize = 1024 * 1024;
```
thomaseizinger, line 23: Perhaps add a comment explaining this constant.

thomaseizinger (commented):
Did a preliminary review, results look great, well done!

I can see that you've spent a fair bit of effort to make the implementation of the algorithm easy to understand. I think it might be worth to improve on that further by decoupling it from the `Stream` struct. We should be able to implement the algorithm as a pure function and simply call it from `Stream` by passing in the required state. That should also allow you to test it more easily using `quickcheck`.

yamux/src/connection.rs
```diff
@@ -378,6 +386,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Result<Stream>> {
         loop {
             if self.socket.poll_ready_unpin(cx).is_ready() {
+                if let Some(frame) = self.rtt.next_ping() {
+                    self.socket.start_send_unpin(frame.into())?;
+                    continue;
+                }
```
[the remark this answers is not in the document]
mxinden: That is a good point.

`next_ping()` starts the timer. Thus I would like the `next_ping` call to be as close to sending the actual frame out as possible.

Where would you put the call to `next_ping()`? Within the `if self.socket.poll_ready` block? Right above the `self.pending_frames.pop_front`?

> You can use `push_front` if you want it to be more timely.

:+1:

yamux/src/connection.rs
```diff
@@ -441,20 +454,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
-            self.pending_frames.push_back(frame.into());
-        }
-
-        let mut stream = self.make_new_outbound_stream(id, self.config.receive_window);
-
-        if extra_credit == 0 {
-            stream.set_flag(stream::Flag::Syn)
-        }
+        let mut stream = self.make_new_outbound_stream(id);
+        stream.set_flag(stream::Flag::Syn);
```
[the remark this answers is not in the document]
mxinden: Good idea. Even better, we can get rid of `Stream::set_flag` entirely, setting the flag on the `new_{in,out}bound` right away.

5d93aa2d4a447049fb94658241a4344ded4b0981

yamux/src/connection/stream.rs
```diff
@@ -200,33 +247,167 @@ impl Stream {
...
+            let mut accumulated_max_stream_windows = self.accumulated_max_stream_windows.lock();
+
+            // Ideally one can just double it:
+            let mut new_max = shared.max_receive_window_size.saturating_mul(2);
+
+            // Then one has to consider the configured stream limit:
+            new_max = cmp::min(
+                new_max,
+                self.config.max_stream_receive_window.unwrap_or(u32::MAX),
+            );
```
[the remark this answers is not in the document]
mxinden: Why do you prefer shadowing over re-assignment? Shadowing bears the risk of creating two variables, with a typo in one. Re-assignment makes that impossible. What I am doing here semantically is re-assignment.

yamux/src/connection/stream.rs
```diff
@@ -200,33 +247,167 @@ impl Stream {
...
+        //
+        // This implementation is heavily influenced by QUIC. See document below for rational on the
+        // above strategy.
+        //
+        // https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit?usp=sharing
+        if self
+            .rtt
+            .get()
+            .map(|rtt| self.last_window_update.elapsed() < rtt * 2)
+            .unwrap_or(false)
```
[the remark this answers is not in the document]
mxinden: Would you prefer the following?

``` rust
        let try_to_auto_tune = self
            .rtt
            .get()
            .map(|rtt| self.last_window_update.elapsed() < rtt * 2)
            .unwrap_or(false);
        if try_to_auto_tune {
```

yamux/src/connection/stream.rs
```diff
@@ -360,15 +541,21 @@ impl AsyncWrite for Stream {
                 log::debug!("{}/{}: can no longer write", self.conn, self.id);
                 return Poll::Ready(Err(self.write_zero_err()));
             }
-            if shared.credit == 0 {
+            if shared.current_send_window_size == 0 {
```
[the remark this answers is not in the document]
mxinden: But neither `window` nor `credit` implies the direction. The `window` on one side is the `credit` of the other. Which one is which? The names don't imply that from the local perspective is inbound and `credit` is outbound.

yamux/src/frame/io.rs
mxinden: :+1: cf76ccf

Note that this pull request does not change the value, but instead makes the previous default value constant. Previously this was set through `max_receive_buffer`. With https://github.com/libp2p/rust-yamux/issues/175 this configuration flag is no longer needed.

yamux/src/connection/stream.rs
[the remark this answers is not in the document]
thomaseizinger: > Shadowing bears the risk of creating two variables, with a typo in one.

I think in most cases, that should yield an "unused variable" lint somewhere.

yamux/src/connection.rs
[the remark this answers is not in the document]
thomaseizinger: Hmm, might as well then leave as is.

yamux/src/connection/stream.rs
```diff
@@ -161,7 +161,7 @@ impl Stream {
             conn,
             config: config.clone(),
             sender,
-            flag: Flag::None,
+            flag: Flag::Ack,
```
thomaseizinger, line 164: Is this change correct? Doesn't this now implicitly ACK all inbound streams?

mxinden: `Stream::flag` is the flag that is piggy-backed onto the next package sent to the remote. This does not in itself trigger a new package being sent containing the acknowledgement. Thus I don't this this will "now implicity ACK all inbound streams". Am I missing something @thomaseizinger?

yamux/src/connection/stream.rs
```diff
@@ -200,33 +247,167 @@ impl Stream {
...
+            );
+            if rtt.is_none() {
+                assert_eq!(
+                    shared.max_receive_window_size, DEFAULT_CREDIT,
+                    "The maximum is only increased iff an rtt measurement is available."
+                );
+            }
+        };
+
+        debug_assert();
```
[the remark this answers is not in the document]
mxinden: Done in https://github.com/libp2p/rust-yamux/pull/176/commits/b9b74db02335ee62660688db406eeaff2ea171d1.

yamux/src/lib.rs
mxinden: I moved these into the respective functions, but still panicking. Would you prefer `Result`s over panicking, or were you just opposed to the separate `check` method?

mxinden:
@thomaseizinger thank you for the feedback! Would you mind taking another look?

yamux/src/connection/stream.rs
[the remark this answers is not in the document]
thomaseizinger: > > If we can remove `mut`, I know that this variable is never re-assigned which IMO is easier to read.
> 
> But you still have to scan the entire function to check whether it is ever shadowed. I argue that this is the same. Only that the `mut` keyword makes it explicit, whereas shadowing is implicit.

With the difference that shadowing is always unconditional because a condition always introduce a nested scope (you can't shadow a variable from within an `if`).  I'd recommend following the rule of least power:

`mut` is more powerful than shadowing, hence its use should be justified. If you don't need `mut`, don't use it! A reader of this code following this rule will otherwise see the `mut`, switch their brain to "careful thinking, we are mutating state here"-mode only to find out that it could have been done without `mut` :)

yamux/src/connection/stream.rs
```diff
@@ -200,33 +247,167 @@ impl Stream {
...
+            // Then one has to consider the configured connection limit:
+            new_max = {
+                let connection_limit: usize = shared.max_receive_window_size as usize +
+                    // the overall configured conneciton limit
+                    self.config.max_connection_receive_window
+                    // minus the minimum amount of window guaranteed to each stream
+                    - self.config.max_num_streams * DEFAULT_CREDIT as usize
+                    // minus the amount of bytes beyond the minimum amount (`DEFAULT_CREDIT`)
+                    // already allocated by this and other streams on the connection.
+                    - *accumulated_max_stream_windows;
```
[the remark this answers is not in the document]
thomaseizinger: The amount of comments tells me that there is a lot going on here so it appears that a dedicated function could abstract this away and free the reader from having to deal with the details when they are just scanning through the function.

In other words, a mathematical statement involving 5 variables seems like something that is worthwhile extracting.

yamux/src/connection/stream.rs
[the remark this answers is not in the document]
thomaseizinger: I don't really mind what we are naming them but I find 4 words too long for this variable, esp. if any related state has 3 out of 4 in common. Just conceptually, this means those 3 words don't add value in differentiating the two from each other.

We never have an "old" or "outdated" window size so `current` is meaningless IMO. `size` doesn't add much either because with in the context of yamux, `send_window` and `receive_window` are expressions of quantities so it is implicit that we are measuring their size.

So maybe just `send_window` and `receive_window`?

yamux/src/connection/stream/flow_control.rs
```diff
@@ -0,0 +1,305 @@
...
+
+        self.assert_invariants(buffer_len);
+
+        return Some(next_window_update);
+    }
+
+    fn assert_invariants(&self, buffer_len: usize) {
+        if !cfg!(debug_assertions) {
+            return;
+        }
```
thomaseizinger, line 134: I find this odd, why not just use `debug_assert` further down? If this is an optimisation, it at least needs a comment.

yamux/src/connection/stream/flow_control.rs
```diff
@@ -0,0 +1,305 @@
...
+use crate::{connection::rtt::Rtt, Config, DEFAULT_CREDIT};
+
+#[derive(Debug)]
+pub(crate) struct FlowController {
+    config: Arc<Config>,
+    last_window_update: Instant,
+    /// See [`Connection::rtt`].
+    rtt: Rtt,
+    /// See [`Connection::accumulated_max_stream_windows`].
+    accumulated_max_stream_windows: Arc<Mutex<usize>>,
```
thomaseizinger, line 14: Could this live outside of `FlowController` instead and you pass a mutable reference into the functions where you need it?

It feels odd that an implementation detail about shared, mutable state leaks into the implementation of this algorithm.

yamux/src/connection/stream/flow_control.rs
```diff
@@ -0,0 +1,305 @@
...
+        debug_assert!(
+            *accumulated_max_stream_windows
+                >= (self.max_receive_window_size - DEFAULT_CREDIT) as usize,
+            "{accumulated_max_stream_windows} {}",
+            self.max_receive_window_size
+        );
+
+        *accumulated_max_stream_windows -= (self.max_receive_window_size - DEFAULT_CREDIT) as usize;
+    }
+}
```
thomaseizinger, line 210: This also doesn't look like it should be part of the algorithm but live in `Stream` instead.

yamux/src/connection/stream/flow_control.rs
thomaseizinger: Same thing with `Arc<Config>` and `Rtt` actually. It would be nice if we could express this algorithm in a more functional way and pass these dependencies in whenever needed. Thoughts?

yamux/src/lib.rs
```diff
@@ -85,36 +82,50 @@ pub struct Config {
 impl Default for Config {
     fn default() -> Self {
         Config {
-            receive_window: DEFAULT_CREDIT,
-            max_buffer_size: 1024 * 1024,
-            max_num_streams: 8192,
+            max_connection_receive_window: Some(1 * 1024 * 1024 * 1024),
```
thomaseizinger, line 85: Perhaps make a 1GB constant?

yamux/src/lib.rs
```diff
@@ -85,36 +82,50 @@ pub struct Config {
...
-        assert!(n >= DEFAULT_CREDIT);
-        self.receive_window = n;
-        self
-    }
+    /// The window of a stream starts at 256 KiB and is increased (auto-tuned) based on the
+    /// connection's round-trip time and the stream's bandwidth (striving for the
+    /// bandwidth-delay-product).
+    ///
+    /// Set to `None` to disable limit, i.e. allow each stream to grow receive window based on
+    /// connection's round-trip time and stream's bandwidth without limit.
```
thomaseizinger, line 104: Shall we add some words around DOS-safety here?

yamux/src/connection/rtt.rs
```diff
@@ -0,0 +1,170 @@
...
+
+#[cfg(test)]
+impl Clone for RttInner {
+    fn clone(&self) -> Self {
+        Self {
+            state: self.state.clone(),
+            rtt: self.rtt.clone(),
+        }
+    }
+}
```
thomaseizinger, line 117: Could be expressed as:

```rust
#[cfg_attr(test, derive(Clone))]
```

yamux/src/connection/rtt.rs
```diff
@@ -0,0 +1,170 @@
...
+const PING_INTERVAL: Duration = Duration::from_secs(10);
+
+#[derive(Clone, Debug)]
+pub(crate) struct Rtt(Arc<Mutex<RttInner>>);
+
+impl Rtt {
+    pub(crate) fn new() -> Self {
+        Self(Arc::new(Mutex::new(RttInner {
+            rtt: None,
+            state: RttState::Initial,
```
thomaseizinger, line 32: Wouldn't `RttState::Waiting { next: Instant::now() }` also be a functionally-equivalent initial state?

yamux/src/connection/rtt.rs
```diff
@@ -0,0 +1,170 @@
...
+        let state = &mut self.0.lock().state;
+
+        match state {
+            RttState::Initial => {}
+            RttState::AwaitingPong { .. } => return None,
+            RttState::Waiting { next } => {
+                if *next > Instant::now() {
+                    return None;
+                }
+            }
```
thomaseizinger, line 46: This doesn't register a waker, hence we won't be measuring RTTs on idle connections, correct? Could we perhaps document this on the struct?

yamux/src/connection/rtt.rs
```diff
@@ -0,0 +1,170 @@
...
+                received_nonce,
+                expected_nonce
+            );
+            return Action::Terminate(Frame::protocol_error());
+        }
+
+        inner.rtt = Some(sent_at.elapsed());
+        log::debug!(
+            "received pong {received_nonce}, estimated round-trip-time {}s",
+            inner.rtt.as_ref().unwrap().as_secs_f64()
```
thomaseizinger, line 81: `Duration` actually has a fairly clever `fmt::Debug` implementation that chooses the correct unit based on its size. Perhaps use that! :)

yamux/src/connection/rtt.rs
```diff
@@ -0,0 +1,170 @@
...
+            }
+            RttState::AwaitingPong { sent_at, nonce } => (sent_at, nonce),
+        };
+
+        if received_nonce != expected_nonce {
+            log::error!(
+                "received pong with {} but expected {}",
+                received_nonce,
+                expected_nonce
+            );
```
thomaseizinger, line 74:
```suggestion
            log::error!("received pong with {received_nonce} but expected {expected_nonce}");
```

yamux/src/connection/stream/flow_control.rs
mxinden: Because `rtt.get()` and `self.accumulated_max_stream_windows.lock()` both take a lock. I would like to not do so in non-debug.

yamux/src/connection/stream/flow_control.rs
mxinden: I would like to bundle all flow-control related logic in one module. Thus I prefer keeping this logic in `flow_control` instead of `stream`.

yamux/src/lib.rs
mxinden: :+1: 8261e12

yamux/src/lib.rs
mxinden: Good idea. Added a section in 7581f76.

yamux/src/connection/rtt.rs
mxinden: :+1: c9c71b6

yamux/src/connection/rtt.rs
mxinden: Oh neat! Thanks @thomaseizinger!

82106b0

yamux/src/connection/rtt.rs
mxinden: Good catch. Indeed. Much better. c3f581f

yamux/src/connection/rtt.rs
mxinden: Perfect! 33c6b05

yamux/src/connection/rtt.rs
mxinden: 68e5e1c

yamux/src/connection/stream.rs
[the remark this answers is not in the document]
mxinden: I can see the argument. That said, I can not think of a function name that is explicit enough to warrant the additional complexity of a function. Will keep as is, unless you feel strongly about it.

thomaseizinger (approved):
Great work @mxinden!

mxinden:
[Tagged](https://github.com/libp2p/rust-yamux/releases/tag/yamux-v0.13.0) and [published](https://crates.io/crates/yamux/0.13.0).

Thank you for your help @thomaseizinger!

mxinden:
Using the [libp2p perf tests](https://github.com/libp2p/test-plans/blob/master/perf/README.md), this change improves **rust-libp2p's TCP/TLS/Yamux single stream performance from ~33 Mbit/s to 1.3 Gbit/s**. See pull request description above for details.

I expect this change to have a large  throughput performance impact on various live networks as well. Thus tagging folks here for visibility.

For those using `yamux` through `libp2p`, see https://github.com/libp2p/rust-libp2p/pull/4970 integrating this change into rust-libp2p `master`, soon to be released as part of the `libp2p` `v0.53` family.

Please reach out in case you have any questions.

- Sigma Prime: @jxs @AgeManning @divagant-martian
- Polkadot SDK: @altonen @bkchr @alindima
- Subspace: @nazar-pc @shamil-gadelshin
- Stormshield: @stormshield-pj50 @stormshield-frb
- Actyx: @rkuhn
- EigerCo: @oblique
- ... please tag others in case I missed someone.

diegomrsantos:
@mxinden great work. Is there a plan to update the Yamux specification to cover this?

mxinden:
Just to avoid confusion, none of this pull request requires changes at the protocol level. In other words, Rust Yamux follows/ed the [libp2p Yamux specification](https://github.com/libp2p/specs/tree/master/yamux) before and after. Round trip time is measured via Yamux's PING message. Window updates are granted via Yamux's WINDOW_UPDATE message.

I am not planning to update the specification any time soon. Though I am happy to review a pull request. Want to give it a shot @diegomrsantos? Suggestions to add:

- Regularly measuring round-trip time via Yamux's PING messages.
- Granting and increasing window based on [QUIC's flow-control algorithm](https://docs.google.com/document/d/1F2YfdDXKpy20WVKJueEf4abn_LVZHhMUMS5gX6Pgjl4/edit#heading=h.j3nq4bn2eaud).

PR #178, opened 2023-11-23
chore: remove `WindowUpdateMode::OnReceive`

mxinden:
Follow up to previous deprecation https://github.com/libp2p/rust-yamux/pull/177.

Fixes https://github.com/libp2p/rust-yamux/issues/175.

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -21,7 +21,7 @@ use crate::{
     error::ConnectionError,
     frame::header::{self, Data, GoAway, Header, Ping, StreamId, Tag, WindowUpdate, CONNECTION_ID},
     frame::{self, Frame},
-    Config, WindowUpdateMode, DEFAULT_CREDIT,
+    Config, DEFAULT_CREDIT,
 };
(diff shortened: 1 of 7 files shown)


mxinden:
Replaced by https://github.com/libp2p/rust-yamux/pull/179 so that I can use stacked pull requests.

yamux/src/connection.rs
```diff
@@ -504,23 +502,13 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
-                        // because the stream will no longer be known.
-                        None
-                    }
+                    // The remote has either still credit or will be given more
+                    // (due to an enqueued window update or because the update
+                    // mode is `OnReceive`) or we already have inbound frames in
+                    // the socket buffer which will be processed later. In any
+                    // case we will reply with an RST in `Connection::on_data`
+                    // because the stream will no longer be known.
+                    None
```
diegomrsantos, line 511: @jxs @mxinden, could you please explain why this is the new correct behavior when the window update mode is always OnRead? It seems to be different from the previous behavior.

PR #180, opened 2023-11-24
chore: move to Rust edition 2021

mxinden:
If I understand correctly, this is a local change only, and thus does not require a changelog entry. Am I correct @thomaseizinger? Any objections?


thomaseizinger (approved):
It potentially raises the MSRV but we don't track that here I think.

PR #190, opened 2024-05-31
fix: wake up readers after setting the state to RecvClosed

### LINKED ISSUE #189: Missed EOF when reading from a stream and the FIN flag is received in a WindowUpdate frame
In case the FIN flag is received in a WindowUpdate frame the code does not wake up readers which can thus miss the stream EOF.

In `connection::Active::on_window_update()` readers are not woken up after setting the state to `RecvClosed`, thus potentially missing the EOF entirely.

kkovaacs:
Readers should be woken up after setting the state to RecvClosed otherwise they potentially miss reading the EOF entirely.

Closes #189


mxinden (approved):
Thanks for debugging! This looks good to me.

@thomaseizinger any objections?

thomaseizinger (approved):

PR #193, opened 2024-06-11
fix: deadlock when sending ping

diegomrsantos:
See https://github.com/libp2p/rust-libp2p/issues/5410 and https://github.com/libp2p/rust-yamux/pull/112

The problem might be that we try to send a ping before reading pending frames. If the receive window is exhausted, no progress can be made.

### DIFF
--- a/test-harness/tests/poll_api.rs
+++ b/test-harness/tests/poll_api.rs
@@ -210,7 +210,7 @@ fn write_deadlock() {
     // We make the message to transmit large enough s.t. the "server"
     // is forced to start writing (i.e. echoing) the bytes before
     // having read the entire payload.
-    let msg = vec![1u8; 1024 * 1024];
+    let msg = vec![1u8; 100024 * 100024];
 
     // We choose a "connection capacity" that is artificially below
     // the size of a receive window. If it were equal or greater,
@@ -219,7 +219,7 @@ fn write_deadlock() {
     // fact that the sum of receive windows of all open streams can easily
     // be larger than the send capacity of the connection at any point in time.
     // Using such a low capacity here therefore yields a more reproducible test.
-    let capacity = 1024;
+    let capacity = 10024;
 
     // Create a bounded channel representing the underlying "connection".
     // Each endpoint gets a name and a bounded capacity for its outbound
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -391,14 +391,6 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
     fn poll(&mut self, cx: &mut Context<'_>) -> Poll<Result<Stream>> {
         loop {
             if self.socket.poll_ready_unpin(cx).is_ready() {
-                // Note `next_ping` does not register a waker and thus if not called regularly (idle
-                // connection) no ping is sent. This is deliberate as an idle connection does not
-                // need RTT measurements to increase its stream receive window.
-                if let Some(frame) = self.rtt.next_ping() {
-                    self.socket.start_send_unpin(frame.into())?;
-                    continue;
-                }
-
                 // Privilege pending `Pong` and `GoAway` `Frame`s
                 // over `Frame`s from the receivers.
                 if let Some(frame) = self
@@ -407,6 +399,14 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                     .or_else(|| self.pending_write_frame.take())
                 {
                     self.socket.start_send_unpin(frame)?;
+                    continue;
+                }
+
+                // Note `next_ping` does not register a waker and thus if not called regularly (idle
+                // connection) no ping is sent. This is deliberate as an idle connection does not
+                // need RTT measurements to increase its stream receive window.
+                if let Some(frame) = self.rtt.next_ping() {
+                    self.socket.start_send_unpin(frame.into())?;
                     continue;
                 }
             }


test-harness/tests/poll_api.rs
```diff
@@ -219,7 +219,7 @@ fn write_deadlock() {
     // fact that the sum of receive windows of all open streams can easily
     // be larger than the send capacity of the connection at any point in time.
     // Using such a low capacity here therefore yields a more reproducible test.
-    let capacity = 1024;
+    let capacity = 10024;
```
diegomrsantos, line 222: Those changes seem to make the test flaky and the deadlock happens sometimes. The test output when it happens is:

```
running 1 test
[2024-06-11T08:40:15Z DEBUG yamux::connection] new connection: 6405a268 (Server)
[2024-06-11T08:40:15Z DEBUG yamux::connection] new connection: d58957af (Client)
[2024-06-11T08:40:15Z DEBUG yamux::connection] d58957af: new outbound (Stream d58957af/1) of (Connection d58957af Client (streams 0))
[2024-06-11T08:40:15Z DEBUG yamux::connection::rtt] sending ping 3047298023
[2024-06-11T08:40:15Z DEBUG yamux::connection::rtt] sending ping 3182731544
[2024-06-11T08:40:15Z DEBUG yamux::connection::rtt] received pong 3047298023, estimated round-trip-time 322.416µs
[2024-06-11T08:40:15Z DEBUG yamux::connection::rtt] received pong 3182731544, estimated round-trip-time 564.167µs
[2024-06-11T08:40:15Z DEBUG yamux::connection::stream::flow_control] old window_max: 0.25 mb, new window_max: 0.5 mb
[2024-06-11T08:40:15Z DEBUG yamux::connection::stream::flow_control] old window_max: 0.25 mb, new window_max: 0.5 mb
[2024-06-11T08:40:15Z DEBUG yamux::connection::stream::flow_control] old window_max: 0.5 mb, new window_max: 1 mb
[2024-06-11T08:40:25Z DEBUG yamux::connection::rtt] sending ping 3157641594
[2024-06-11T08:40:25Z DEBUG yamux::connection::rtt] sending ping 1136718128
[2024-06-11T08:40:25Z DEBUG yamux::connection::rtt] received pong 3157641594, estimated round-trip-time 459.708µs
[2024-06-11T08:40:25Z DEBUG yamux::connection::rtt] received pong 1136718128, estimated round-trip-time 145.75µs
[2024-06-11T08:40:25Z DEBUG yamux::connection::stream::flow_control] old window_max: 0.5 mb, new window_max: 1 mb
[2024-06-11T08:40:35Z DEBUG yamux::connection::rtt] sending ping 1970012150
[2024-06-11T08:40:35Z DEBUG yamux::connection::rtt] sending ping 2852812461
[2024-06-11T08:40:35Z DEBUG yamux::connection::rtt] received pong 1970012150, estimated round-trip-time 180.5µs
[2024-06-11T08:40:35Z DEBUG yamux::connection::rtt] received pong 2852812461, estimated round-trip-time 87.458µs
[2024-06-11T08:40:45Z DEBUG yamux::connection::rtt] sending ping 2531287629
[2024-06-11T08:40:45Z DEBUG yamux::connection::rtt] sending ping 1628284672
test write_deadlock has been running for over 60 seconds
```

Would it also be possible to achieve the same using smaller numbers?

yamux/src/connection.rs
```diff
@@ -409,6 +401,14 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                     self.socket.start_send_unpin(frame)?;
                     continue;
                 }
+
```
diegomrsantos, line 404: Changing this order seem to fix the issue.

dariusc93: Does it deadlock on yamux test too? If not, maybe we could add a test for this too?

diegomrsantos: Could you please elaborate? I understand this test was designed specifically to address this type of deadlock.

dariusc93: I actually missed the test that was there so my comment can be ignored :)

test-harness/tests/poll_api.rs
diegomrsantos: Using 

```rust
let msg = vec![1u8; 30 * 1024 * 1024];
let capacity = 102;
pub const DEFAULT_CREDIT: u32 = 25600;
const PING_INTERVAL: Duration = Duration::from_millis(500);
```
I can reproduce it at least once when running 5 times. But not when capacity is 100. Maybe it can give some idea about the cause.

diegomrsantos: How can I override `DEFAULT_CREDIT` and `PING_INTERVAL` for this test? I changed the values where they are defined as I didn't know how to override them in the test.

dariusc93:
CC @mxinden

mxinden:
> The problem might be that we try to send a ping before reading pending frames. If the receive window is exhausted, no progress can be made.

Can you expand how that would lead to a dead-lock? Once the local node has sent the ping, it would then proceed to reading pending frames. Also who's receive window?

diegomrsantos:
> > The problem might be that we try to send a ping before reading pending frames. If the receive window is exhausted, no progress can be made.
> 
> Can you expand how that would lead to a dead-lock? Once the local node has sent the ping, it would then proceed to reading pending frames. Also who's receive window?

I don't know much about the internals, but my theory is that when the remote receive window has been exhausted, we try to send a ping before reading the frame which increases our credit again. If both peers do this at the same time, the deadlock happens as shown here https://github.com/libp2p/rust-yamux/pull/193#discussion_r1634544258. I couldn't reproduce the deadlock anymore after reading the frame before sending the ping. I apologize, but this PR is mostly empirical, I don't have much knowledge about the system yet to make a better analysis. I hope it can be somehow helpful.

dariusc93:
any update on this?

CC @diegomrsantos

test-harness/tests/poll_api.rs
```diff
@@ -210,7 +210,7 @@ fn write_deadlock() {
     // We make the message to transmit large enough s.t. the "server"
     // is forced to start writing (i.e. echoing) the bytes before
     // having read the entire payload.
-    let msg = vec![1u8; 1024 * 1024];
+    let msg = vec![1u8; 100024 * 100024];
```
jxs, line 213: can we maintain the previous values?
I feel only Increasing the flakiness rate isn't much valuable

jxs (approved):
Hi @diegomrsantos! Thanks for this, LGTM, left a comment.
 
> Can you expand how that would lead to a dead-lock? Once the local node has sent the ping, it would then proceed to reading pending frames. Also who's receive window?

https://github.com/libp2p/rust-yamux/blob/70632689b5e6103e068f683ec2bc87d5e8fe54a3/yamux/src/connection.rs#L393-L412

Hi @mxinden it doesn't proceed to read the pending frames because it won't enter the `if` bracket as the `socket` is not ready, the `Ping` has not been sent.

By switching the order we privilege sending `Frame`s that clear `pending_read_frame` and allow us to make progress reading. 
We were already privileging clearing `pending_read_frame`s over `pending_write_frame`s, this change makes it so that sending a `Ping` has a lower priority than the those.
If you agree we can merge this PR.

Thanks

mxinden:
I am not sure I fully understand the deadlock. It is at the TCP flow-control level?



> We were already privileging clearing `pending_read_frame`s over `pending_write_frame`s, this change makes it so that sending a `Ping` has a lower priority than the those.

Prioritizing `Ping` frames is an optimization, leading to better RTT measurements only.

The optimization should not be prioritized over correct behavior. In other words, if this can indeed lead to a dead-lock, then yes, your change to the order makes sense.

test-harness/tests/poll_api.rs
diegomrsantos: I didn't mean to merge those changes. Do you happen to know how to test this PR properly?

test-harness/tests/poll_api.rs
jxs: how would we benefit from changing those values? `DEFAULT_CREDIT` is part of the spec and `PING_INTERVAL` is not user configurable.

jxs (commented):
> Prioritizing Ping frames is an optimization, leading to better RTT measurements only.

>The optimization should not be prioritized over correct behavior. In other words, if this can >indeed lead to a dead-lock, then yes, your change to the order makes sense.

Thanks for elaborating Max, I also spent a couple minutes today trying to reproduce the issue but I could only do it changing `DEFAULT_CREDIT`and `PING_INTERVAL` which don't say much as the former is part of the spec and the latter is not user configurable.

just increasing the payload on `write_deadlock` test I couldn't reproduce the aforementioned issue.
@indietyp do you have a reproducible case for this issue?
I suggest we wait for more conclusive data

indietyp:
Sadly I don't have a reliable reproducible case for this, I tried or quite a while.

I just found that this would semi-reliably happen when using yamux 1.3 - so it's _some_ kind of edge case.

I'll try again maybe I'll find a reproducible case, tho I am not filled with confidence 😅

jxs:
@dariusc93 was able to found a reproducible example that I pushed to https://github.com/jxs/rust-libp2p/tree/yamux-deadlock,
to reproduce it, one has to checkout that branch, `yamux-deadlock` and run the `stream` example, both the listener instance and the connecting instance.
Contrary to the example on `master` the one on `yamux-deadlock` eventually stops echoing.

diegomrsantos:
> @dariusc93 was able to found a reproducible example that I pushed to https://github.com/jxs/rust-libp2p/tree/yamux-deadlock, to reproduce it, one has to checkout that branch, `yamux-deadlock` and run the `stream` example, both the listener instance and the connecting instance. Contrary to the example on `master` the one on `yamux-deadlock` eventually stops echoing.

I ran it on macOS and I don't see any echo. Could you confirm this behavior is caused by the deadlock? I'm not entirely sure I'm testing it correctly, but I still can't see any echo after applying the fix in this PR.

diegomrsantos:
The potential solution proposed here doesn't fix the issue, as the cause doesn't seem related to the ping.

The code here https://github.com/jxs/rust-libp2p/blob/yamux-deadlock/examples/stream/src/main.rs successfully reproduces the issue. It happens when the client streams around 500 KB of data to a server which echoes it back to the client.
After a deeper investigation, @jxs and I found the cause seems to be related to flow control, specifically these lines https://github.com/diegomrsantos/rust-yamux/blob/master/yamux/src/connection/stream/flow_control.rs#L52:L54.

The server decides no window update needs to be sent, as the condition is satisfied, but right after neither the client nor the server has more credits left. 

It's still unclear what that's the case, but it should probably be addressed in a new PR. Thanks to everyone who helped to clarify the issue.

Teebor-Choka:
Is there any relevant PR tackling this ATM?

jxs:
looked more into this, my example hangs because the sender exhausts the default window size and doesn't read the window update from the remote which is expected behavior described by Diego on https://github.com/libp2p/rust-yamux/pull/193#issuecomment-2200816518.

What we could do is mimic go implementation and implement the keep alive and write timeout.  Feeling like closing this one and continuing the discussion on https://github.com/libp2p/rust-libp2p/issues/5410

PR #194, opened 2024-10-15
Fix closing pending frames

turuslan:
When yamux connection is closing it should:
- Send pending frames.
- Send `GoAway` frame.
- Close underlying connection.

But test showed that pending frames from `stream_receivers` and `GoAway` frame were not sent.

Commits introducing this bug:
- https://github.com/libp2p/rust-yamux/commit/61003c53cec1b5b37f8c05231e44efea0077ad07 replaced `VecDeque` with `Option`,
  which caused loss of pending frames,
  because only last pending frame was stored in `Option`
- https://github.com/libp2p/rust-yamux/commit/297e9eba1f30ded3a2b4bb46c8bd8cf2fb4471db reordered sending and collecting pending frames,
  trying to fix loss of pending frames.
- https://github.com/libp2p/rust-yamux/commit/286728026d8ec15cb4120028d3fabdbd35ebf3ca reverted https://github.com/libp2p/rust-yamux/commit/61003c53cec1b5b37f8c05231e44efea0077ad07 `Option` back to `VecDeque`,
  so all pending frames are stored again.
- However https://github.com/libp2p/rust-yamux/commit/297e9eba1f30ded3a2b4bb46c8bd8cf2fb4471db was not reverted,
  so pending frames are collected,
  but not sent.

This PR provides test to reproduce it, and reverts https://github.com/libp2p/rust-yamux/commit/297e9eba1f30ded3a2b4bb46c8bd8cf2fb4471db and https://github.com/libp2p/rust-yamux/commit/af8f6935c1aca23d8efc59743d665d7bc5ffe414 to fix it.

### DIFF
--- a/yamux/src/connection/closing.rs
+++ b/yamux/src/connection/closing.rs
@@ -98,3 +98,102 @@ enum State {
     FlushingPendingFrames,
     ClosingSocket,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::future::poll_fn;
+    use futures::FutureExt;
+
+    struct Socket {
+        written: Vec<u8>,
+        closed: bool,
+    }
+    impl AsyncRead for Socket {
+        fn poll_read(
+            self: Pin<&mut Self>,
+            _: &mut Context<'_>,
+            _: &mut [u8],
+        ) -> Poll<std::io::Result<usize>> {
+            todo!()
+        }
+    }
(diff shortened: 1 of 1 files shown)


jxs (approved):
Thanks for this! Yeah I overlooked that `DrainingStreamReceiver` pushes frames back to the  end of the queue.
Will release a patch version with this fix later on

PR #195, opened 2024-10-23
prepare 0.13.4

jxs:


dariusc93:
I wonder if we should try to resolve https://github.com/libp2p/rust-yamux/pull/193 before publishing the patch release. Thoughts?

guillaumemichel (approved):
LGTM, thanks @jxs!

@dariusc93 we can always publish a new patch release when that PR is merged. No need to wait for it

PR #196, opened 2024-10-31
Implement Connection::get_ref to retrieve a reference to the socket

stormshield-fabs:
This lets users of `Connection` immutably access the socket they provided.

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -88,6 +88,11 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Connection<T> {
         }
     }
 
+    /// Retrieve a reference to the socket.
+    pub fn get_ref(&self) -> Option<&T> {
+        self.inner.get_ref()
+    }
+
     /// Poll for a new outbound stream.
     ///
     /// This function will fail if the current state does not allow opening new outbound streams.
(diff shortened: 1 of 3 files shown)


jxs (commented):
Hi! And thanks for this, can you give more context on why you need to `AsyncRead` and `AsyncWrite` on the internal connection?

stormshield-fabs:
Hello,
We don't need to use these traits on the internal connection, but we need to directly access attributes that cannot be put behind a synchronization primitive for performance reasons.

Now that I'm re-reading this, maybe the implementations of `get_ref` should be in different `impl` blocs without the trait bounds?

jxs:
> We don't need to use these traits on the internal connection, but we need to directly access attributes that cannot be put behind a synchronization primitive for performance reasons.

But how do you access those attributes if you are returning a `T`? 
Can you disclose which attributes are?

jxs:
> The precise use case is for metrics: the `AsyncRead` and `AsyncWrite` methods on our socket type increment `u64`s stored in the structure.
> 
> ```rust
> let socket = Socket { metric: 0, ... };
> let connection = yamux::Connection::new(socket, ...)
> // calls to `poll_new_outbound`, `poll_next_inbound`...
> let metric = connection.get_ref().unwrap().metric;
> ```

thanks for elaborating, but if you are calling `AsyncRead` and `AsyncWrite` on the inner connection you are messing the internal behaviour, can you explain what metrics would you like to record? I think it's better to just do as we do in `rust-libp2p` and offer explicit API's for that

stormshield-fabs:
We never call `AsyncRead` nor `AsyncWrite` on the inner connection, but I see how the proposed patch allows this undesirable behavior.

The metrics are two simple counters for the total written and read bytes on the connection. Can you show point me to the explicit APIs in `rust-libp2p` you mentioned?

jxs:
> The metrics are two simple counters for the total written and read bytes on the connection. Can you link me to the explicit APIs you mentioned in `rust-libp2p`?

`rust-libp2p` has its whole [metrics](https://docs.rs/libp2p/latest/libp2p/metrics/struct.Metrics.html) crate, which we don't need in this case, we can just expose methods that return the total bytes read and written on `Connection` which you can then record metrics, wdyt?

stormshield-fabs:
Yes that would work.

Let's close this PR and I'll submit another with the changes you suggest (except if you'd prefer to take care of it!).

jxs:
no no, go ahead!
Thanks :rocket:

Issue #170, opened 2023-07-20
Refactor `stream::Shared` to use internal mutability


thomaseizinger:
Currently, we hold a `stream::Shared` within the stream and one in `Active`. This is shared by being wrapped in `Arc<Mutex>`.

It would be slightly more ergonomic for both usage sites if the `Arc<Mutex>` would be inside `stream::Shared`.

akaladarshi:
Hey @mxinden 

Is this issue still available for pickup?

dariusc93:
Yes it is @akaladarshi :)

akaladarshi:
@dariusc93 Can you assign it to me?.

Also just to get the idea: We need to put `Arc<Mutex>` inside [stream::Shared](https://github.com/libp2p/rust-yamux/blob/3f9e8c3649ba4215937c9d6344f729ff97c747c2/yamux/src/connection/stream.rs#L93), so we don't need to use `Arc<Mutex<Shared>>` everywhere else.

PR #197, opened 2025-01-26
refactor: Shared to use internal mutability

akaladarshi:
Fixes: #170 

This PR changes:
- Now the `Shared` struct internally handles the mutex.
- `Shared` struct now has `inner: Arc<Mutex<SharedInner>>`

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -282,7 +282,7 @@ struct Active<T> {
     socket: Fuse<frame::Io<T>>,
     next_id: u32,
 
-    streams: IntMap<StreamId, Arc<Mutex<stream::Shared>>>,
+    streams: IntMap<StreamId, stream::Shared>,
     stream_receivers: SelectAll<TaggedStream<StreamId, mpsc::Receiver<StreamCommand>>>,
     no_streams_waker: Option<Waker>,
 
@@ -507,9 +507,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         let s = self.streams.remove(&stream_id).expect("stream not found");
 
         log::trace!("{}: removing dropped stream {}", self.id, stream_id);
-        let frame = {
-            let mut shared = s.lock();
-            let frame = match shared.update_state(self.id, stream_id, State::Closed) {
+        let frame = s.with_mut(|inner| {
+            let frame = match inner.update_state(self.id, stream_id, State::Closed) {
                 // The stream was dropped without calling `poll_close`.
                 // We reset the stream to inform the remote of the closure.
                 State::Open { .. } => {
@@ -541,14 +540,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
                 // remote end has already done so in the past.
                 State::Closed => None,
             };
-            if let Some(w) = shared.reader.take() {
+            if let Some(w) = inner.reader.take() {
                 w.wake()
             }
-            if let Some(w) = shared.writer.take() {
+            if let Some(w) = inner.writer.take() {
                 w.wake()
             }
+
             frame
-        };
+        });
         frame.map(Into::into)
     }
 
@@ -565,10 +565,8 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             && matches!(frame.header().tag(), Tag::Data | Tag::WindowUpdate)
         {
             let id = frame.header().stream_id();
-            if let Some(stream) = self.streams.get(&id) {
-                stream
-                    .lock()
-                    .update_state(self.id, id, State::Open { acknowledged: true });
+            if let Some(shared) = self.streams.get(&id) {
+                shared.update_state(self.id, id, State::Open { acknowledged: true });
             }
             if let Some(waker) = self.new_outbound_stream_waker.take() {
                 waker.wake();
@@ -590,14 +588,15 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
         if frame.header().flags().contains(header::RST) {
             // stream reset
             if let Some(s) = self.streams.get_mut(&stream_id) {
-                let mut shared = s.lock();
-                shared.update_state(self.id, stream_id, State::Closed);
-                if let Some(w) = shared.reader.take() {
-                    w.wake()
-                }
-                if let Some(w) = shared.writer.take() {
-                    w.wake()
-                }
+                s.with_mut(|inner| {
+                    inner.update_state(self.id, stream_id, State::Closed);
+                    if let Some(w) = inner.reader.take() {
+                        w.wake()
+                    }
+                    if let Some(w) = inner.writer.take() {
+                        w.wake()
+                    }
+                });
             }
             return Action::None;
         }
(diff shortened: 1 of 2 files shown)


yamux/src/connection/stream.rs
```diff
@@ -316,37 +317,47 @@ impl AsyncRead for Stream {
         }
 
         // Copy data from stream buffer.
-        let mut shared = self.shared();
         let mut n = 0;
-        while let Some(chunk) = shared.buffer.front_mut() {
-            if chunk.is_empty() {
-                shared.buffer.pop();
-                continue;
+        let can_read = self.shared.with_mut(|inner| {
```
elenaf9, line 321: Can't we just return here directly the `Poll` result?
That way we could move the whole bottom part of the function as it was inside this closure.

yamux/src/connection/stream.rs
```diff
@@ -393,8 +415,9 @@ impl AsyncWrite for Stream {
         // a) to be consistent with outbound streams
         // b) to correctly test our behaviour around timing of when ACKs are sent. See `ack_timing.rs` test.
         if frame.header().flags().contains(ACK) {
-            self.shared()
-                .update_state(self.conn, self.id, State::Open { acknowledged: true });
+            self.shared.with_mut(|inner| {
+                inner.update_state(self.conn, self.id, State::Open { acknowledged: true });
+            });
```
elenaf9, line 420:
```suggestion
            self.shared.update_state(self.conn, self.id, State::Open { acknowledged: true });
```

Isn't there a `update_state` convenience function on `Shared`?

yamux/src/connection/stream.rs
```diff
@@ -429,30 +452,90 @@ impl AsyncWrite for Stream {
...
+        self.with_mut(|inner| {
+            inner.reader = waker;
+        });
+    }
+
+    pub fn update_state(&self, cid: connection::Id, sid: StreamId, next: State) -> State {
+        self.with_mut(|inner| inner.update_state(cid, sid, next))
+    }
+
+    pub fn with_mut<F, R>(&self, f: F) -> R
```
elenaf9, line 512: `parking_lot::Mutex` doesn't support re-entrant locking, thus a nested call to this would cause a deadlock. 
To avoid this, the function should take `&mut self`.

yamux/src/connection/stream.rs
```diff
@@ -429,30 +452,90 @@ impl AsyncWrite for Stream {
         self.sender
             .start_send(cmd)
             .map_err(|_| self.write_zero_err())?;
-        self.shared()
-            .update_state(self.conn, self.id, State::SendClosed);
+        self.shared.with_mut(|inner| {
+            inner.update_state(self.conn, self.id, State::SendClosed);
+        });
```
elenaf9, line 457: Same here: i think we can use directly `Shared::update_state`?

elenaf9 (commented):
Thank you for your PR @akaladarshi! 

Some comments.

yamux/src/connection/stream.rs
akaladarshi: Thanks for pointing this out!

I need a suggestion regarding the design choices in the code snippet below. We are frequently using `Pin<&mut Self> ` in `connection.rs`. So if  we change `fn with
_mut` to take `&mut self` then, we would run into borrow checkers issue.

So to fix it, is the below snippet a good choice or we could do something like take a `mutexGuard` on sharedInner or is there any better?.

Here’s the code for reference (stream.rs -> fn poll_next):

```rust
        let conn_id = self.conn.clone();
        let stream_id = self.id.clone();
        let shared = &mut self.shared;
        let poll_state = shared.with_mut(|inner| {
            if let Some(bytes) = inner.buffer.pop() {
                let off = bytes.offset();
                let mut vec = bytes.into_vec();
                if off != 0 {
                    // This should generally not happen when the stream is used only as
                    // a `futures::stream::Stream` since the whole point of this impl is
                    // to consume chunks atomically. It may perhaps happen when mixing
                    // this impl and the `AsyncRead` one.
                    log::debug!(
                        "{}/{}: chunk has been partially consumed",
                        conn_id,
                        stream_id,
                    );
                    vec = vec.split_off(off)
                }
                return Poll::Ready(Some(Ok(Packet(vec))));
            }

            // Buffer is empty, let's check if we can expect to read more data.
            if !inner.state.can_read() {
                log::debug!("{}/{}: eof", conn_id, stream_id);
                return Poll::Ready(None); // stream has been reset
            }

            // Since we have no more data at this point, we want to be woken up
            // by the connection when more becomes available for us.
            inner.reader = Some(cx.waker().clone());
            Poll::Pending
        });
```

yamux/src/connection.rs
```diff
@@ -628,35 +627,40 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
-                shared.consume_receive_window(frame.body_len());
-                shared.buffer.push(frame.into_body());
+                stream.shared().with_mut(|inner| {
+                    if is_finish {
+                        inner.update_state(self.id, stream_id, State::RecvClosed);
+                    }
+                    inner.consume_receive_window(frame.body_len());
+                    inner.buffer.push(frame.into_body());
+                })
             }
```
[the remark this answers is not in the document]
akaladarshi: Reverted the changes to use `self.shared.lock()`.

yamux/src/connection.rs
```diff
@@ -628,35 +627,40 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
...
+                } else {
+                    if is_finish {
+                        inner.update_state(self.id, stream_id, State::RecvClosed);
+                    }
+                    inner.consume_receive_window(frame.body_len());
+                    inner.buffer.push(frame.into_body());
+                    if let Some(w) = inner.reader.take() {
+                        w.wake()
+                    }
+                    Action::None
```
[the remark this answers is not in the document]
akaladarshi: Done.

yamux/src/connection/stream.rs
```diff
@@ -262,9 +265,7 @@ impl futures::stream::Stream for Stream {
             Poll::Pending => {}
         }
 
-        let mut shared = self.shared();
-
-        if let Some(bytes) = shared.buffer.pop() {
+        if let Some(bytes) = self.shared.pop_buffer() {
```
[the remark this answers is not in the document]
akaladarshi: Removed it, now using single lock.

yamux/src/connection/stream.rs
akaladarshi: It's now using simple `lock` instead of `with_mut`, so no need for this anymore.

yamux/src/connection/stream.rs
akaladarshi: Done.

yamux/src/connection/stream.rs
akaladarshi: Updated the `with_mut` to use `&mut`.

I added another `lock()` function, which will allows us to use simple lock wherever it's necessary and also keeping the `with_mut` so we can use whichever we need.

yamux/src/connection/stream.rs
akaladarshi: Done.

yamux/src/connection/stream.rs
elenaf9: > I need a suggestion regarding the design choices in the code snippet below. We are frequently using `Pin<&mut Self> `in `connection.rs.` So if we change `fn with_mut` to take `&mut self` then, we would run into borrow checkers issue.

You could use `Pin::get_mut` and then destructure `Self` to access the distinct fields, which would satisfy the borrow checker:
```rs
        let Self {
            conn, shared, id, ..
        } = self.get_mut();
        shared.with_mut(|inner| {
            inner.update_state(*conn, *id, State::SendClosed);
        });
```

> I added another `lock()` function, which will allows us to use simple lock wherever it's necessary and also keeping the `with_mut` so we can use whichever we need.

`lock` also needs to take `&mut self` for the same reasons.
Generally, I am not a fan of having both `lock` and `with_mut` since they are just two ways of doing the same thing. I am okay with either of them, but we should decide for one.

akaladarshi: Ohh okay so essential same as storing seperately as my example suggest but using `get_mut`.

Then I think we will go with the `with_mut` as it is more dynamic implementation.

akaladarshi:
@elenaf9 Thanks for the review 🙏. 

 I have removed the `lock` and directly using `with_mut`, just one small change I have refactored Stream `write_zero_err` fn to be a static as it was just for creating error and it was causing few issues with borrow checkers in `poll_write`.

yamux/src/connection/stream.rs
```diff
@@ -360,54 +362,65 @@ impl AsyncWrite for Stream {
...
+            .left();
+
+        stream.add_flag(frame.header_mut());
+
+        log::trace!("{}/{}: write {} bytes", stream.conn, stream.id, n);
 
-        // technically, the frame hasn't been sent yet on the wire but from the perspective of this data structure, we've queued the frame for sending
-        // We are tracking this information:
-        // a) to be consistent with outbound streams
-        // b) to correctly test our behaviour around timing of when ACKs are sent. See `ack_timing.rs` test.
```
elenaf9, line 394: I don't think this comment should be removed?

elenaf9 (approved):
One comment, rest LGTM. Thank you for the follow-ups @akaladarshi!

cc @jxs: think this change is an improvement because it hides the mutex-locking logic behind the `Shared` type and prevents deadlocks that would otherwise been possible when trying to locking the Mutex re-entrantly. Are you okay with merging?

yamux/src/connection/stream.rs
akaladarshi: Done.

PR #199, opened 2025-04-02
fix: avoid possible poisoned connection when closing

sdbondi:
The connection state can become `Poisoned` if poll_close returns an error. Although an implementer should never poll again after an error, the connection state should also never be `Poisoned` (`unreachable!`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.13.5
+
+- Fix possible poisoned connection when closing. See [PR 199](https://github.com/libp2p/rust-yamux/pull/199).
+
 # 0.13.4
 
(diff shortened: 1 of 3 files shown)


elenaf9 (approved):
Good catch, thanks! 
Could you please add a CHANGELOG entry and bump the `yamux` patch version?

jxs (approved):
Thanks! :heart:  Do you need a release?

sdbondi:
@jxs Sorry for the late reply and thanks for the merge - no rush, we have fixed our code so that we do not poll after errors

dizda:
Thanks for providing this fix, I think a release would makes sense.

jxs:
published `yamux` [`0'.13.5`](https://crates.io/crates/yamux)

PR #219, opened 2026-02-25
feat: add configurable yamux close timeout for unread remote peers

jxs:
This introduces a configurable timeout for `yamux` connection shutdown, so `poll_close` can complete even when the remote peer stops reading from the underlying transport.
When the timeout elapses during yamux close, the implementation stops waiting on yamux-level frame draining and force-closes at the yamux state-machine level.

In backpressured scenarios (e.g. remote stops reading), graceful yamux shutdown may otherwise wait indefinitely while trying to flush pending frames. This change gives users a bounded close path for those situations.

**Important**: this timeout only affects yamux protocol-level shutdown behavior.  
It does **not** replace transport/socket-level close semantics. For TCP-level behavior (e.g. FIN/RST/linger behavior), users should still configure socket options such as `SO_LINGER` as appropriate for their environment.

### DIFF
--- a/test-harness/tests/poll_api.rs
+++ b/test-harness/tests/poll_api.rs
@@ -329,3 +329,37 @@ fn close_through_drop_of_stream_propagates_to_remote() {
     })
     .unwrap();
 }
+
+#[test]
+fn close_timeout_force_closes_when_remote_stops_reading() {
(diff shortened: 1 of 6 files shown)


ackintosh (approved):
Looks good to me!

PR #229, opened 2026-06-18
feat: replace rand-based IDs and nonce with atomic counters

jxs:
Replace the runtime usages of rand::random() with atomic counters. 
The connection identifier was only used for debug logging and had no protocol significance, so a sequential counter from `AtomicU32` should suffice. 

The same applies to the ping identifier, matching the approach used in the Go yamux reference implementation where a sequentially incremented `uint32` under a `mutex` is used, [1](https://github.com/hashicorp/yamux/blob/master/session.go#L333-L338). 

Move `rand` to dev-dependencies, it's now only required in test code to fill frame bodies with random bytes.

### DIFF
--- a/yamux/src/connection.rs
+++ b/yamux/src/connection.rs
@@ -17,6 +17,8 @@ mod closing;
 mod rtt;
 mod stream;
 
+use std::sync::atomic::{AtomicU32, Ordering};
+
 use crate::tagged_stream::TaggedStream;
 use crate::{
     error::ConnectionError,
@@ -37,6 +39,9 @@ use std::{fmt, sync::Arc, task::Poll};
 
 pub use stream::{Packet, State, Stream};
 
+/// Next connection identifier, used for debug logging.
+static NEXT_ID: AtomicU32 = AtomicU32::new(1);
+
 /// How the connection is used.
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
 pub enum Mode {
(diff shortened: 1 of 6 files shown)


yamux/src/frame.rs
```diff
@@ -133,8 +133,8 @@ impl Frame<WindowUpdate> {
 }
 
 impl Frame<Ping> {
-    pub fn ping(nonce: u32) -> Self {
```
rmonteverde70-create, line 136: sa

rmonteverde70-create (approved):

yamux/src/connection/rtt.rs
```diff
@@ -45,34 +51,34 @@ impl Rtt {
             }
         }
 
-        let nonce = rand::random();
+        let id = NEXT_PING.fetch_add(1, Ordering::SeqCst);
```
dknopik: Why `Ordering::SeqCst` instead of `Ordering::Relaxed`?

yamux/src/connection.rs
```diff
@@ -48,14 +53,14 @@ pub enum Mode {
...
 #[derive(Clone, Copy)]
 pub(crate) struct Id(u32);
 
 impl Id {
-    /// Create a random connection ID.
-    pub(crate) fn random() -> Self {
-        Id(rand::random())
+    /// Create a new connection ID.
+    pub(crate) fn next() -> Self {
+        Id(NEXT_ID.fetch_add(1, Ordering::SeqCst))
```
dknopik: dito

yamux/src/connection/rtt.rs
jxs: good question, `SeqCst` is my go to default, but you are right, `Relaxed` suffices here. Thanks Daniel. updated!

dknopik (approved):
lgtm

gitToki (approved):
LGTM!

0xMushow:
LGTM, small divergences from the Go implementation though, but doesn't impact functionality:

- Go seems to have a per-session counter, rather than a global veriable, since it's within the "Session" struct. If we want to fully match it we could put it in `RttInner`
- Go impl starts at 0, not 1.

jxs:
> LGTM, small divergences from the Go implementation though, but doesn't impact functionality:
> 
> * Go seems to have a per-session counter, rather than a global veriable, since it's within the "Session" struct. If we want to fully match it we could put it in `RttInner`
> * Go impl starts at 0, not 1.

Thanks! Aplied those changes!

dknopik (approved):
LGTM

PR #233, opened 2026-07-25
Fix deadlock when a stream is split

### LINKED ISSUE #232: Yamux deadlock when splitting stream
https://github.com/stormshield-pj50/yamux_race is a small example that reproduces a deadlock in Yamux when a stream is split.

The program launches a server and a client. The client establishes a TCP connection to the server and opens a single Yamux stream. Both the server and the client then split that stream into a read half and a write half, and each side spawns two asynchronous tasks: one that continuously writes to the stream and another that continuously reads from it.

After the writer has performed ten writes, the deadlock occurs. The issue appears to arise when the internal MPSC sender on the Yamux stream becomes full. At that point the writer task is never woken up again because the reader registers another waker when it sends a WindowUpdate frame on the MPSC channel. The Yamux internal MPSC sender is shared by both the reader and the writer tasks. The MPSC implementation of the future crate wakes up only the last waker. So the writer task waker is definitely lost.

The deadlock pattern is reproduced by this simple example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5fb34c6f773d51b0fd12bba6a2b5d798 

Tokio references this issue : https://github.com/tokio-rs/tls/issues/40

Tudyx:
Fixes https://github.com/libp2p/rust-yamux/issues/232

As described in https://github.com/libp2p/rust-yamux/issues/232 the `Read` and the `Write` part of a `Stream` shared the same `Sender` and hence the same waker slot. This can lead to a "lost wake-up" bug when the read and the write are polled from 2 different tasks, for instance after a `futures::AsyncReadExt::split`. The existing tests didn't catch it because they always drive both halves from a single task, and a single task means a single waker.

The proposed fix is to have one sender handle for the writer and one for the reader, both to the same channel.

I wrote two tests, one that really pins the specific issue, and one more global that reproduces the reported symptom end to end.

I did a first CI run to prove that the regression tests fail (you can see the result [here](https://github.com/Tudyx/rust-yamux/actions/runs/30159379964/job/89682086504?pr=1)), and then pushed the fix.

This PR was written primarily by Claude Code, this PR description by me. I carefully reviewed each line of code, and I stand behind it.
I didn't see an AI contribution policy, let me know if that's ok.


mcthesw (approved):
I ran into the same lost-wakeup issue with yamux 0.14.0 through a Tokio IO adapter, with the read and write halves running in separate tasks.
I tested this PR against our reproducer and confirmed that it fixes the issue. Thanks for working on this!

elenaf9 (approved):
LGTM! Thanks!

yamux/src/connection/stream.rs
```diff
@@ -95,7 +95,18 @@ pub struct Stream {
...
     sender: mpsc::Sender<StreamCommand>,
+    /// Sends the window updates of the reading side of this stream to the connection.
+    ///
+    /// This is a second handle to the same channel as [`Stream::sender`]. A
+    /// [`mpsc::Sender`] has room for exactly one waker, which it overwrites whenever the
+    /// channel is full. Sharing a single handle between the two halves of a split stream
+    /// therefore makes the half polled last discard the waker of the other one, which is then
+    /// never woken again.
+    ///
+    /// See https://github.com/libp2p/rust-yamux/issues/232.
```
elenaf9, line 108: No need to link the issue. The explanation is enough.

```suggestion
```

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.14.1
+
+- Fix a deadlock when a `Stream` is split and its two halves are driven by separate tasks.
+  Both halves used to share a single channel handle to the connection, whose only waker slot
+  the half polled last would overwrite, so the other half was never woken again.
+  See [issue 232](https://github.com/libp2p/rust-yamux/issues/232).
```
elenaf9, line 6:
```suggestion
  See [PR 233](https://github.com/libp2p/rust-yamux/pull/233).
```

Tudyx:
Thanks for the review

elenaf9 (approved):
LGTM, thanks.

jxs (approved):
LGTM!

